PR #2, opened 2020-10-26
add bridge control logic

yaocw2020:
- bridge operation
   - Ensure: ensure bridge has been existed with configuration as expected.
   - BorrowAddr: borrow address from specify nic.
   - ReturnAddr: return address to specify nic before delete bridge or reset nic.

- bridge vlan controller logic
   - OnChange:
     1. Check whether the bridge exist
     - Yes，do nothing
     - No,  add new one
     2. Check whether the master of nic is bridge
     - Yes，do nothing
     - No, make bridge return addresses to configured nic, lend addresses to bridge
     3. Check whether the bridge has at least one address
     - Yes, do nothing
     - No, borrow the addresses from nic
   - OnDelete: return address and delete bridge.
- nad controller: watch nad and update bridge vlan


guangbochen:
@yaocw2020 can u please add the commit message and split the commit into 2, 1st one for code changes, and 2nd one for vendor update.

pkg/network/bridge_vlan_controller.go
```diff
@@ -39,35 +46,109 @@ func Register(ctx context.Context, apply apply.Apply, setting harvcontroller.Set
...
+	bytes, err := json.Marshal(networkSetting)
+	if err != nil {
+		return nil, fmt.Errorf("marshal failed, error: %w, networkSetting: %+v", err, networkSetting)
 	}
 
-	// TODO, bridge VLAN setting is not suppose to be removed, therefore we can add a backup setting on delete?
+	setting.Value = string(bytes)
 
-	return setting, nil
+	return c.settingClient.Update(setting)
```
guangbochen, line 96: slim this method:
1. do and only do one thing in this method, encodes and return the `networkSetting` JSON string 
2. move the update logic to the onChange method and need to do deepCopy when updating the object.

pkg/network/bridge_vlan_controller.go
```diff
@@ -39,35 +46,109 @@ func Register(ctx context.Context, apply apply.Apply, setting harvcontroller.Set
...
+		return nil, err
 	}
-	logrus.Printf("harvester network setting configured to: %s", setting.Value)
 
-	// TODO, config/re-config the bridge
-	// 1. convert setting to networSetting struct, return error if is invalid
+	// update configured nic
+	networkSetting.ConfiguredNIC = networkSetting.NIC
+	return c.updateNetworkSetting(networkSetting, setting)
+}
```
guangbochen, line 61: check `updateNetworkSetting` comments.

pkg/network/bridge_vlan_controller.go
```diff
@@ -39,35 +46,109 @@ func Register(ctx context.Context, apply apply.Apply, setting harvcontroller.Set
...
 
-	// 2. if has configured-bridge, reset the bridge and configured NIC first
+func (c *BridgeVLANController) OnRemove(key string, setting *harvesterv1.Setting) (*harvesterv1.Setting, error) {
+	// TODO, bridge VLAN setting is not suppose to be removed, therefore we can add a backup setting on delete?
+	networkSetting, err := NewNetworkSetting(setting)
+	if err != nil {
+		return nil, fmt.Errorf("get network setting failed, error;: %w, setting: %+v", err, setting)
+	}
```
guangbochen, line 68: ditto, except don't need to check `setting.DeletionTimestamp` is nil

pkg/network/bridge_vlan_controller.go
```diff
@@ -67,7 +79,81 @@ func (c *BridgeVLANController) OnRemove(key string, setting *harvesterv1.Setting
...
+		}
+		return nil
+	}
+
+	// Check whether the master of nic is bridge
+	// - Yes，do nothing
+	// - No, make bridge return addresses to configured nic, lend addresses to bridge, return.
+	if nic.Attrs().MasterIndex != c.bridge.Index {
+		if err := c.bridge.ReturnAddr(configuredNIC); err != nil {
+			klog.Errorf("Return address failed, error: %v, NIC: %s", err, setting.ConfiguredNIC)
```
gitlawr, line 134: Why do we proceed when ReturnAddr fails? (including the L124)

guangbochen (commented):
we should store the `IP` and its `route` in the annotation to avoid it get lost in certain circumstances.

pkg/bridge/bridge.go
```diff
@@ -0,0 +1,170 @@
...
+// Ensure bridge
+// set promiscuous mod default
+func (br *Bridge) Ensure() error {
+	if err := netlink.LinkAdd(br); err != nil && err != syscall.EEXIST {
+		return fmt.Errorf("add bridge failed, error: %w, bridge: %v", err, br)
+	}
+
+	// Re-fetch link to read all attributes and if it already existed,
+	// ensure it's really a bridge with similar configuration
+	if err := br.fetchByName(); err != nil {
```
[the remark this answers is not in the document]
yaocw2020: package netlink has not supported to update vlan_filtering attribute of bridge, so we leave a todo.

pkg/network/bridge_vlan_controller.go
yaocw2020: > Why do we proceed when ReturnAddr fails? (including the L124)

In some case, ReturnAddr will fail, for example, previous configured nic is deleted manually. But I think the bridge configuration process needs to go on.

guangbochen (approved):
LGTM

pkg/controller/nad/nad_controller.go
```diff
@@ -0,0 +1,109 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	l, err := c.getLink()
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO delete previous vlan id when update nad
```
gitlawr, line 56: What happens before this is implemented?

go.mod
```diff
@@ -15,9 +15,9 @@ replace (
 
 require (
 	github.com/dgrijalva/jwt-go v3.2.1-0.20200107013213-dc14462fd587+incompatible // indirect
-	github.com/j-keck/arping v1.0.1
-	github.com/rancher/harvester v0.0.2-0.20201026052331-8cbe186706f1
 	github.com/rancher/lasso v0.0.0-20200515155337-a34e1e26ad91
+	github.com/k8snetworkplumbingwg/network-attachment-definition-client v0.0.0-20200331171230-d50e42f2b669
```
gitlawr, line 18: Conventionally, we put go.mod/go.sum changes in the vendor update commit

pkg/controller/vlan/bridge_vlan_controller.go
```diff
@@ -0,0 +1,226 @@
...
+	if setting == nil {
+		return nil, nil
+	}
+
+	if setting.Value == "" || key != NetworkSettingName {
+		return setting, nil
+	}
+
+	klog.Info("harvester network setting has been deleted.")
+	networkSetting, err := DecodeNetworkSettings(setting.Value)
```
gitlawr, line 101: Can we return early if ConfiguredNIC==NIC?

pkg/controller/nad/nad_controller.go
yaocw2020: Currently, we can not get old vlan id.

pkg/controller/vlan/bridge_vlan_controller.go
yaocw2020: In some case, for example nodes reboot and network configurations are lost, we still need to configure network even if ConfiguredNIC==NIC.

gitlawr (approved):
LGTM

Issue #7, opened 2020-11-16
Default pvid can not be configured by nad controller


yaocw2020:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
Create nad with default pvid 1

**Result:**
NAD controller change bridge vlan, resulting in network unreachable. The arp reply accepted by ens160 (configured nic) will not be sent to harvester-br0.
![image](https://user-images.githubusercontent.com/7421463/99237476-7bc6c780-2833-11eb-84dd-b5469d671a79.png)
![image](https://user-images.githubusercontent.com/7421463/99237717-d52ef680-2833-11eb-98dd-5a17298fc1a7.png)
![image](https://user-images.githubusercontent.com/7421463/99238312-8766be00-2834-11eb-9522-d20e9a73b5ad.png)
![image](https://user-images.githubusercontent.com/7421463/99238405-a6655000-2834-11eb-89a0-90a90b1bb903.png)
![image](https://user-images.githubusercontent.com/7421463/99238503-c432b500-2834-11eb-929c-16af8ace8f93.png)

yaocw2020:
The correct default pvid configuration should be like that:
```
port	vlan ids
ens160	 1 PVID Egress Untagged
```

PR #8, opened 2020-11-16
Forbidden nad controller configures default pvid

yaocw2020:
**issue**: #7


pkg/bridge/link.go
```diff
@@ -6,6 +6,8 @@ import (
 	"github.com/vishvananda/netlink"
 )
 
+const DefaultPVID = uint16(1)
```
guangbochen, line 9: small case here

gitlawr (approved):
LGTM

PR #12, opened 2020-11-18
Update network setting when it is empty

guangbochen:
**Problem:**
When the network-setting is reset to an empty value the configured status is still showing as `true`

**Solution:**
update the configured status according to the input NIC value

https://github.com/rancher/harvester-network-controller/issues/11

### DIFF
--- a/pkg/controller/vlan/bridge_vlan_controller.go
+++ b/pkg/controller/vlan/bridge_vlan_controller.go
@@ -82,8 +82,13 @@ func (c *BridgeVLANController) OnChange(key string, setting *harvesterv1.Setting
 		return nil, fmt.Errorf("encode network settings failed, error: %w, networksettings: %+v", err, networkSetting)
 	}
 
-	harvesterv1.SettingConfigured.True(settingCopy)
+	if networkSetting.NIC == "" {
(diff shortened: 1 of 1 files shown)


pkg/controller/vlan/bridge_vlan_controller.go
```diff
@@ -82,8 +82,14 @@ func (c *BridgeVLANController) OnChange(key string, setting *harvesterv1.Setting
 		return nil, fmt.Errorf("encode network settings failed, error: %w, networksettings: %+v", err, networkSetting)
 	}
 
-	harvesterv1.SettingConfigured.True(settingCopy)
-	harvesterv1.SettingConfigured.Reason(settingCopy, "")
+	if networkSetting.NIC == "" {
+		harvesterv1.SettingConfigured.False(settingCopy)
+		harvesterv1.SettingConfigured.Reason(settingCopy, "")
```
yaocw2020, line 87: Why not put this line out of if-else block?

guangbochen: updated

yaocw2020:
LGTM

PR #15, opened 2020-12-02
Fix drone can't find release

guangbochen:
Fix drone can't find release https://drone-publish.rancher.io/rancher/harvester-network-controller/12/1/3


guangbochen:
close as this is caused by incorrect drone git_secret key

PR #25, opened 2021-03-16
Optimize VLAN network routes

yaocw2020:
- Replace `netlink.RouteReplace` with `netlink.RouteAppend`  and `netlink.RouteDel`

  > With`netlink.RouteReplace` to add routes, It may replace the existed routes related with other NICs. 

- Replace `GetVlanWithNic` with `GetVlan`

   > We have got configured NIC from nodenetwork status as a parameter of `GetVlanWithNic` before. However, if fail to update status, what recorded in nodenetwork status is inconsistent with the actual situation. At this moment, vlan object returned by `GetVlanWithNic` is no a real VLAN. With it to tear down VLAN network, it will bring misconfiguration.

- Optimize `vlan.Setup` and `vlan.Teaddown`

   > Adjust configuration orders to reduce negative impact on the original network. For example, add routes and IPv4 address to bridge before attaching the physical NIC with bridge.

- Delete `vlan.afterDelBridge`

   > It's too unstable to delete this function because it's hard to recover the routes rightly. Bridge deletion is a low frequency event and we will reconsider to implement automatic restoration of the bridge if there is a customer demand.


gitlawr:
Can you elaborate on the reason for the change

yaocw2020:
> Can you elaborate on the reason for the change

Sure

pkg/controller/agent/nodenetwork/controller.go
```diff
@@ -196,20 +199,17 @@ func (h Handler) getNICFromStatus(nn *networkv1alpha1.NodeNetwork) string {
...
-	if err != nil {
-		return fmt.Errorf("new vlan with nic %s failed, error: %w", configuredNic, err)
+	if errors.As(err, &netlink.LinkNotFoundError{}) || errors.As(err, &vlan.SlaveNotFound{}) {
+		klog.Infof("ignore link not found error, details: %+v", err)
+		return nil
 	}
 
-	if err := v.Repeal(); err != nil {
+	if err := v.Teardown(); err != nil {
 		return fmt.Errorf("repeal vlan failed, error: %w", err)
```
gitlawr, line 213: Update the error message accordingly

pkg/network/iface/bridge.go
```diff
@@ -52,101 +52,38 @@ func (br *Bridge) Ensure() error {
...
+// Keep the bridge's IPv4 addresses are the same with the slave
+func (br *Bridge) SyncIPv4Addr(slave IFace) error {
+	delList := relativeComplement(slave.Addr(), br.addr)
+	addList := relativeComplement(br.addr, slave.Addr())
 	for _, addr := range delList {
 		klog.Infof("delete address: %+v", addr)
 		if err := netlink.AddrDel(br.bridge, &addr); err != nil {
 			return fmt.Errorf("could not add address, error: %w, link: %s, addr: %+v", err, br.bridge.Name, addr)
 		}
+		klog.Infof("delete IPv4 address %+v", addr)
```
gitlawr, line 67: This duplicate with L63

pkg/network/vlan/vlan.go
```diff
@@ -39,53 +39,112 @@ func NewVlan(helper *network.Helper) *Vlan {
...
 	if err := v.bridge.Ensure(); err != nil {
 		return fmt.Errorf("ensure bridge %s failed, error: %w", v.bridge.Name(), err)
 	}
-
 	l, err := iface.GetLink(nic)
 	if err != nil {
-		return err
+		return fmt.Errorf("get NIC %s failed, error: %w", nic, err)
+	}
+	// TODO judge power off?
```
gitlawr, line 90: remove this

pkg/network/vlan/vlan.go
```diff
@@ -39,53 +39,112 @@ func NewVlan(helper *network.Helper) *Vlan {
...
+
+	klog.Info("setup VLAN network successfully")
+	return nil
+}
+
+// Note: It's required to call function GetVlanWithNic before tearing down VLAN.
+func (v *Vlan) Teardown() error {
+	klog.Info("start to tear down VLAN network")
+	if v.nic == nil {
+		return fmt.Errorf("vlan network has't attached a NIC")
```
gitlawr, line 123: Fix typo

pkg/network/vlan/vlan.go
```diff
@@ -251,15 +243,37 @@ func (v *Vlan) afterLinkDown(update netlink.LinkUpdate) {
 			Reason:    "LinkDown",
 			Message:   fmt.Sprintf("Link %s has been down atypically and the controller will try to set it up", update.Link.Attrs().Name),
 		}
-
 		if v.helper != nil && v.helper.EventSender != nil {
 			if err := v.helper.EventSender(event, v.Type()); err != nil {
 				klog.Errorf("recorde event failed, error: %s", err.Error())
+				return
```
gitlawr, line 249: What happens when it returns here? Will it do another try to recover routes?

pkg/network/helper.go
```diff
@@ -2,7 +2,6 @@ package network
 
 type Helper struct {
 	EventSender func(event *Event, networkType string) error
-	Resetter    func(networkType string) error
 }
```
gitlawr, line 5: The Helper type seems unnecessary. can it be something like
```
type EventSender interface{
  func(event *Event, networkType string) error
}
```

pkg/network/vlan/vlan.go
yaocw2020: No

pkg/network/helper.go
yaocw2020: Yes, it seems that interface is better than structure. I reserve the name `helper` and reimplement it as an interface instead of structure.

pkg/network/vlan/vlan.go
yaocw2020: I will delete `return` sentence to let it continue to try to set the down link up or sync IPv4 addresses.

pkg/controller/agent/nad/controller.go
```diff
@@ -59,20 +56,15 @@ func (h Handler) OnChange(key string, nad *nadv1.NetworkAttachmentDefinition) (*
 
 	// TODO delete previous vlan id when update nad
 
-	name := os.Getenv(common.KeyNodeName) + "-" + string(networkv1alpha1.NetworkTypeVLAN)
-	nn, err := h.nodeNetworkCache.Get(common.Namespace, name)
-	if err != nil && !apierrors.IsNotFound(err) {
-		return nil, fmt.Errorf("get node network %s failed, error: %w", name, err)
+	v, err := vlan.GetVlan()
+	if err != nil && !errors.As(err, &netlink.LinkNotFoundError{}) && !errors.As(err, &vlan.SlaveNotFound{}) {
```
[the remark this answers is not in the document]
yaocw2020: We use errors.As() here in order to distinguish the `LinkNotFoundError` and `SlaveNotFoundError`.  These two errors show that VLAN network has not been configured.

pkg/network/helper.go
gitlawr: Help is too generic as an interface name so I suggest we change that

pkg/controller/agent/nad/controller.go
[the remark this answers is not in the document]
gitlawr: Isn't it `errors.Is()`?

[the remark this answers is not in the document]
yaocw2020: errors.Is() is used to match error, including the error message inside. Here what we to compare is only the error type.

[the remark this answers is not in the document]
yaocw2020: Attach annotation of `netlink.LinkNotFoundError`
> // LinkNotFoundError wraps the various not found errors when
// getting/reading links. This is intended for better error
// handling by dependent code so that "not found error" can
// be distinguished from other errors

pkg/network/vlan/vlan.go
```diff
@@ -17,7 +17,7 @@ type Vlan struct {
 	bridge *iface.Bridge
 	nic    *iface.Link
 	status *network.Status
-	helper *network.Helper
+	helper network.EventSender
```
gitlawr, line 20: Can we update the field name

pkg/network/iface/link.go
```diff
@@ -161,25 +175,36 @@ func (l *Link) UnsetRules4DHCP() error {
...
+		err := netlink.RouteAppend(&route)
+		if err != nil && !errors.Is(err, syscall.EEXIST) {
+			return fmt.Errorf("append route failed, route: %+v, error: %w", route, err)
+		}
+		if err == nil {
+			klog.Infof("append route: %+v", route)
+		}
+		if errors.Is(err, syscall.EEXIST) {
+			klog.Infof("ignore existed route: %+v", route)
+		}
```
gitlawr, line 196: let's simplify these ifs

guangbochen (approved):
LGTM

PR #26, opened 2021-03-30
Initialize nodenetwork status

yaocw2020:
Fix: https://github.com/rancher/harvester/issues/611
- Remove subresource status
- Initialize nodenetwork status

### DIFF
--- a/manifests/crds/network.harvester.cattle.io_nodenetworks.yaml
+++ b/manifests/crds/network.harvester.cattle.io_nodenetworks.yaml
@@ -166,8 +166,7 @@ spec:
         type: object
     served: true
     storage: true
-    subresources:
-      status: {}
+    subresources: {}
 status:
   acceptedNames:
     kind: ""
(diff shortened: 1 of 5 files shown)


pkg/controller/common/nodenetwork.go
```diff
@@ -17,6 +17,7 @@ const (
 	Namespace        = "harvester-system"
 	KeyNodeName      = "NODENAME"
 	KeyLabelNodeName = "network.harvester.cattle.io/nodename"
+	initStatusMsg    = "initialize"
```
gitlawr, line 20: Is this what we show in the UI? We should make it more readable.
showing a message like this could be confusing: `node1: initialize`

manifests/crds/network.harvester.cattle.io_nodenetworks.yaml
```diff
@@ -166,8 +166,7 @@ spec:
         type: object
     served: true
     storage: true
-    subresources:
-      status: {}
+    subresources: {}
```
gitlawr, line 169: Why do we remove the subresource?

yaocw2020: With subresources, we can not initialize status when creating nodenetwork.

pkg/controller/common/nodenetwork.go
yaocw2020: Change message as `initializing network configuration` according to what suggested by guangbo.

guangbochen (approved):
LGTM

gitlawr (approved):

PR #28, opened 2021-04-21
keep default route after address changed

yaocw2020:
Issue: https://github.com/harvester/harvester/issues/709
Solution:
- Modify the read-write mutex in the monitor package to mutex to ensure that only one `netlink` event is handled at a time.
- Delete routes of harvester-br0 before synchronizing address.
- Watch routes of the slave NIC and transform them to harvester-br0.

### DIFF
--- a/pkg/network/monitor/monitor.go
+++ b/pkg/network/monitor/monitor.go
@@ -16,7 +16,7 @@ import (
 
 type Monitor struct {
 	handlers map[int]Handler
-	mutex    sync.RWMutex
+	mutex    sync.Mutex
 
 	done chan struct{}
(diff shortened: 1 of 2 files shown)


guangbochen:
@yaocw2020 can u please leave a comment on the issue of how to verify this issue with reproducible steps.

yaocw2020:
> @yaocw2020 can u please leave a comment on the issue of how to verify this issue with reproducible steps.

I updated the steps in the related bug issue.

gitlawr (approved):
LGTM

guangbochen (approved):

PR #29, opened 2021-04-25
Fix repo reference 

yaocw2020:
Issue: https://github.com/harvester/harvester/issues/727
Solution: 
- replace rancher reference with harvester
- update Kubernetes version to keep the same with harvester


.drone.yml
```diff
@@ -45,7 +45,7 @@ steps:
       dockerfile: package/Dockerfile
       password:
         from_secret: docker_password
-      repo: "rancher/harvester-network-controller"
+      repo: "harvester/harvester-network-controller"
```
gitlawr, line 48: Keep docker images unchanged

gitlawr (changes requested):

pkg/apis/network.harvesterhci.io/v1beta1/zz_generated_register.go
```diff
@@ -21,7 +21,7 @@ limitations under the License.
 package v1beta1
 
 import (
-	network "github.com/rancher/harvester-network-controller/pkg/apis/network.harvesterhci.io"
+	network "github.com/harvester/harvester-network-controller/pkg/apis/network.harvesterhci.io"
```
gitlawr, line 24: move to the generated commit

scripts/package
```diff
@@ -9,7 +9,7 @@ SUFFIX="-${ARCH}"
 cd $(dirname $0)/..
 
 TAG=${TAG:-${VERSION}${SUFFIX}}
-REPO=${REPO:-rancher}
+REPO=${REPO:-harvester}
```
gitlawr, line 12: Keep docker images unchanged

pkg/apis/network.harvesterhci.io/v1beta1/zz_generated_register.go
yaocw2020: Sad, I always miss this file.

gitlawr (approved):
LGTM

PR #31, opened 2021-04-27
Fix network  vendor and use cluster-admin

futuretea:
After bump harvester, harvester-network-controller unable to register controllers
and also need use wrangler generated api instead of wranger-api to avoid panic

**Related issue**
harvester/harvester#775


guangbochen (approved):
LGTM

gitlawr (approved):
LGTM

PR #33, opened 2021-04-29
Ignore the same IP address

yaocw2020:
Issue: https://github.com/harvester/harvester/issues/709
Solution: If the IP address is not changed, we don't need to synchronize the IP address.


guangbochen (approved):
LGTM

gitlawr (approved):
LGTM

PR #39, opened 2021-07-21
Use sub-interface

yaocw2020:
To use a macvlan interface rather than the NIC itself if it's used by the management network.

issue: https://github.com/harvester/harvester/issues/980


guangbochen (approved):
LGTM

gitlawr (approved):
LGTM

PR #42, opened 2021-08-25
Ensure ip forward for VLAN

yaocw2020:
issue: https://github.com/harvester/harvester/issues/841


pkg/network/iface/link.go
```diff
@@ -203,6 +208,63 @@ func (l *Link) unsetRules4DHCP() error {
...
+	}
+
+	for _, rule := range rules {
+		if strings.HasPrefix(rule, "-P "+chainForward) {
+			if strings.Fields(rule)[2] == "ACCEPT" {
+				return nil
+			}
+			break
+		}
+	}
```
gitlawr, line 239: Use `ipt.Exists()` for the check

yaocw2020: What we check is the policy, not a rule.  The function `ipt.Exists` doesn't work in this case.

gitlawr (approved):
LGTM

guangbochen (approved):
LGTM

PR #43, opened 2021-09-07
Support vip

yaocw2020:
Add vip controller to support vip.

issue: https://github.com/harvester/harvester/issues/633

### DIFF
--- a/manifests/deployment.yaml
+++ b/manifests/deployment.yaml
@@ -58,11 +58,12 @@ spec:
       containers:
         - image: rancher/harvester-network-controller:master-head
           imagePullPolicy: IfNotPresent
-          name: network
+          name: harvester-manager-network-controller
           command:
             - harvester-network-controller
           args:
             - manager
+            - --enable-vip-controller
           resources:
             limits:
               cpu: 500m
--- a/pkg/controller/common/nodenetwork.go
+++ b/pkg/controller/common/nodenetwork.go
@@ -50,6 +50,7 @@ func NewNodeNetworkFromNode(node *corev1.Node, networkType networkv1.NetworkType
 	// initialize status
 	networkv1.NodeNetworkReady.SetStatusBool(nn, false)
 	networkv1.NodeNetworkReady.Message(nn, initStatusMsg)
+	networkv1.NodeNetworkRemoved.SetStatusBool(nn, false)
 
 	switch networkType {
 	case networkv1.NetworkTypeVLAN:
--- a/pkg/controller/manager/setup.go
+++ b/pkg/controller/manager/setup.go
@@ -1,12 +1,24 @@
 package manager
 
 import (
+	"github.com/urfave/cli"
+
 	"github.com/harvester/harvester-network-controller/pkg/config"
 	"github.com/harvester/harvester-network-controller/pkg/controller/manager/clusternetwork"
 	"github.com/harvester/harvester-network-controller/pkg/controller/manager/node"
+	"github.com/harvester/harvester-network-controller/pkg/controller/manager/vip"
 )
 
-var RegisterFuncList = []config.RegisterFunc{
+var registerFuncList = []config.RegisterFunc{
 	clusternetwork.Register,
 	node.Register,
+}
+
+func GetRegisterFuncList(c *cli.Context) []config.RegisterFunc {
+	enableVipController := c.Bool("enable-vip-controller")
+	if enableVipController {
+		registerFuncList = append(registerFuncList, vip.Register)
+	}
+
+	return registerFuncList
 }
--- a/pkg/controller/manager/vip/controller.go
+++ b/pkg/controller/manager/vip/controller.go
@@ -1,0 +1,148 @@
+package vip
+
+import (
+	"context"
+	"fmt"
+
+	ctlappsv1 "github.com/rancher/wrangler/pkg/generated/controllers/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/klog"
+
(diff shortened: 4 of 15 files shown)


pkg/controller/common/nodenetwork.go
```diff
@@ -14,6 +14,7 @@ import (
 )
 
 const (
+	Namespace        = "harvester-system"
```
guangbochen, line 17: why don't we get ns from https://github.com/harvester/network-controller-harvester/blob/master/main.go#L47

pkg/controller/manager/setup.go
```diff
@@ -4,9 +4,11 @@ import (
 	"github.com/harvester/harvester-network-controller/pkg/config"
 	"github.com/harvester/harvester-network-controller/pkg/controller/manager/clusternetwork"
 	"github.com/harvester/harvester-network-controller/pkg/controller/manager/node"
+	"github.com/harvester/harvester-network-controller/pkg/controller/manager/vip"
 )
 
 var RegisterFuncList = []config.RegisterFunc{
 	clusternetwork.Register,
 	node.Register,
+	vip.Register,
```
guangbochen, line 13: prefer to add a common option to either enable or disable the VIP integration via https://github.com/harvester/network-controller-harvester/blob/master/main.go#L33

pkg/controller/manager/vip/controller.go
```diff
@@ -0,0 +1,137 @@
...
+
+	var vipInterface string
+	var containerIndex, envIndex int
+	for i, c := range ds.Spec.Template.Spec.Containers {
+		if c.Name == vipContainerName {
+			containerIndex = i
+			for j, env := range c.Env {
+				if env.Name == vipInterfaceEnvName {
+					envIndex = j
+					vipInterface = env.Value
```
guangbochen, line 115: should break once found it

pkg/controller/manager/vip/controller.go
```diff
@@ -0,0 +1,137 @@
...
+	if nn.Spec.Type != networkv1.NetworkTypeVLAN {
+		return nn, nil
+	}
+
+	for _, con := range nn.Status.Conditions {
+		if con.Type == networkv1.NodeNetworkRemoved {
+			if con.Status == corev1.ConditionTrue {
+				return nn, h.updateVipInterface(vlan.BridgeName)
+			}
+			return nil, fmt.Errorf("the vlan network hasn't removed")
```
guangbochen, line 91: should return this at the end of func.

pkg/controller/common/nodenetwork.go
yaocw2020: It's the namespace where kube-vip deploys.

guangbochen: /s/Namespace/KubeVipNamespace/

manifests/deployment.yaml
```diff
@@ -58,11 +58,12 @@ spec:
       containers:
         - image: rancher/harvester-network-controller:master-head
           imagePullPolicy: IfNotPresent
-          name: network
+          name: harvester-manager-network-controller
```
guangbochen, line 61: Nit: shorter is better since this is the container name

manifests/deployment.yaml
```diff
@@ -58,11 +58,12 @@ spec:
...
           command:
             - harvester-network-controller
           args:
             - manager
+            - --enable-vip-controller
```
guangbochen, line 66: default to false, therefore this is useless, you can enable it on the harvester side.

pkg/controller/manager/setup.go
```diff
@@ -1,12 +1,25 @@
...
 	"github.com/harvester/harvester-network-controller/pkg/config"
 	"github.com/harvester/harvester-network-controller/pkg/controller/manager/clusternetwork"
 	"github.com/harvester/harvester-network-controller/pkg/controller/manager/node"
+	"github.com/harvester/harvester-network-controller/pkg/controller/manager/vip"
 )
 
-var RegisterFuncList = []config.RegisterFunc{
-	clusternetwork.Register,
-	node.Register,
+var registerMap = map[string]config.RegisterFunc{
```
gitlawr, line 12: I think we don't need to change RegisterFuncList to a map. We can append `vip.Register` to the funcList when `--enable-vip-controller` is set

pkg/controller/manager/vip/controller.go
```diff
@@ -0,0 +1,139 @@
...
+		}
+	}
+
+	if vipInterface != unavailableNIC {
+		return nil
+	}
+
+	dsCopy := ds.DeepCopy()
+	if vipInterface == vlan.BridgeName {
+		dsCopy.Spec.Template.Spec.Containers[containerIndex].Env[envIndex].Value = defaultVipInterface
```
gitlawr, line 129: We are using eth0 as a fixed default. Should it be configurable?

pkg/controller/manager/vip/controller.go
```diff
@@ -0,0 +1,139 @@
...
+	} else {
+		dsCopy.Spec.Template.Spec.Containers[containerIndex].Env[envIndex].Value = vlan.BridgeName
+	}
+
+	if _, err := h.dsCtr.Update(dsCopy); err != nil {
```
gitlawr, line 134: Will this env change get overridden on helm upgrade?

yaocw2020: Only in the case of upgrading kube-vip

gitlawr (approved):
LGTM

guangbochen (approved):
LGTM

PR #45, opened 2021-09-22
Only delete macvlan when setting bridge as master

yaocw2020:
Generally, the parent index of the calico veth interface is 3, which is equal to the second NIC. When using the second NIC as the network interface for the VLAN network, the interface whose parent index equals the index of the second NIC will be deleted. The deletion is aimed to clear the macvlan normally, not the veth. We distinguish the interface type to avoid deleting the calico veth atypically.

issue: https://github.com/harvester/harvester/issues/1259

### DIFF
--- a/pkg/network/iface/link.go
+++ b/pkg/network/iface/link.go
@@ -108,17 +108,18 @@ func (l *Link) DelBridgeVlan(vid uint16) error {
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


gitlawr (approved):
LGTM

guangbochen (approved):

PR #46, opened 2021-09-26
Initialize nodenetwork

yaocw2020:
- Initialize node network even if the VLAN network has not been enabled.
- Optimize the network interface status.

https://github.com/harvester/harvester/issues/1273


yaocw2020:
TODO: update the helm chart of the Harvester network controller.

pkg/controller/agent/nodenetwork/controller.go
```diff
@@ -220,14 +221,9 @@ func (h Handler) removeVlanNetwork(nn *networkv1.NodeNetwork) error {
...
-	// update removed condition
-	nnCopy := nn.DeepCopy()
-	networkv1.NodeNetworkRemoved.SetStatusBool(nnCopy, true)
-	if _, err := h.nodeNetworkCtr.Update(nnCopy); err != nil {
-		return err
-	}
-
-	return nil
+	return h.updateReadyStatus(nn, network.Status{
+		Condition: network.Condition{Normal: false, Message: "The VLAN network has been tear down"},
```
guangbochen, line 225: \s\tear down\teardown

pkg/controller/agent/nodenetwork/controller.go
```diff
@@ -249,11 +245,11 @@ func (h Handler) updateReadyStatus(nn *networkv1.NodeNetwork, status network.Sta
 	}
 
 	// get all physical NICs
-	nics, err := h.getNICs()
+	nis, err := h.getNetworkInterfaces()
```
guangbochen, line 248: using `nic` and `nics` instead of `ni` of the following changes

pkg/controller/agent/nodenetwork/controller.go
```diff
@@ -314,31 +310,40 @@ func makeLinkStatus(link iface.IFace) *networkv1.LinkStatus {
...
 	}
 
-	mgmtNICIndex := h.mgmtNetwork.NIC().Index()
+	mgmtNiIndex := h.mgmtNetwork.NIC().Index()
 
-	nics := []networkv1.NIC{}
+	var vlanNiIndex int
+	v, err := vlan.GetVlan()
+	if err == nil {
+		vlanNiIndex = v.NIC().Index()
```
guangbochen, line 324: why would user be able to get VLAN info when there is a error?

pkg/controller/agent/nodenetwork/controller.go
```diff
@@ -314,31 +310,40 @@ func makeLinkStatus(link iface.IFace) *networkv1.LinkStatus {
...
+	}
+
+	nis := []networkv1.NetworkInterface{}
```
guangbochen, line 327: use make to initialise instead

pkg/controller/common/nodenetwork.go
```diff
@@ -121,3 +120,36 @@ func CreateAllNodeNetworkIfNotExist(networkType networkv1.NetworkType, cn *netwo
 
 	return nil
 }
+
+// if cluster network is open, set the default NIC for all node networks.
+// if cluster network is close, clear the NIC specification of all node networks.
+func SetNICForAllNodeNetworks(enable bool, nic string, cache ctlnetworkv1.NodeNetworkCache, client ctlnetworkv1.NodeNetworkClient) error {
```
guangbochen, line 126: define a common struct instead of passing the cache and client directly.

yaocw2020: We passed the cache and client directly for all functions in this common file. Reserve this style to avoid affecting other functions in this PR and I will submit another PR to optimize it after the 0.3 release. What do you think?

guangbochen: sure

pkg/apis/network.harvesterhci.io/v1beta1/nodenetwork.go
```diff
@@ -33,7 +33,7 @@ type NodeNetworkSpec struct {
 	Type NetworkType `json:"type,omitempty"`
 
 	// +optional
-	NIC string `json:"nic,omitempty"`
+	NetworkInterface string `json:"nic,omitempty"`
```
gitlawr, line 36: Generally, we should keep the json key and the field name consistent. If we don't change the json key, we should keep the filed name unchanged

yaocw2020: It's temporary. We consider the compatibility between single cluster UI and Rancher UI. Rancher UI can not be modified now while single cluster UI will add this new feature. https://github.com/harvester/harvester/issues/1273

guangbochen (approved):
LGTM

gitlawr (approved):
LGTM

PR #47, opened 2021-11-02
Modify NIC to network interface

yaocw2020:
Related issue: https://github.com/harvester/harvester/issues/1273


guangbochen:
can u test the upgrade case, from v0.3 to the master-head.

PR #48, opened 2021-11-05
Add routes from hosts to vlan network

yaocw2020:
Solution:
- Add the annotation `network.harvesterhci.io/layer3-network` to nad CR. There are two ways to get the layer 3 network information.
   - `auto`: Create a job to set up a network helper pod to get the CIDR and gateway through the DHCP discovery.
   - `manual`: Users should configure the CIDR and gateway manually. It's the default option.
- Ping the gateway of the VLAN network recorded in the annotation periodically to check the connectivity from host to the VLAN network. The controller will record the connectivity in the annotation as well.
- Add the route to reach the addresses in the CIDR via the default gateway on every host.

Related issue: 
https://github.com/harvester/harvester/issues/1476
https://github.com/harvester/harvester/issues/1474
https://github.com/harvester/harvester/issues/1059
https://github.com/harvester/harvester/issues/1424

- [x] Modify CI configuration
- [x] file an EIO ticket


pkg/utils/nadconf.go
```diff
@@ -0,0 +1,74 @@
...
+	Auto   Mode = "auto"
+	Manual Mode = "manual"
+)
+
+type NadSelectedNetworks []nadv1.NetworkSelectionElement
+
+type Layer3NetworkConf struct {
+	Mode         Mode         `json:"mode"`
+	CIDR         string       `json:"cidr"`
+	Gateway      string       `json:"gateway"`
```
guangbochen, line 35: can we add parse CIDR and IP validations?

pkg/utils/nadconf.go
```diff
@@ -0,0 +1,74 @@
...
+	Connectivity Connectivity `json:"connectivity"`
```
guangbochen, line 36: when the mode is auto, can the user specify the DHCP IP, also should we allow users to config the DNS as an optional option here?

pkg/controller/manager/nad/controller.go
```diff
@@ -0,0 +1,334 @@
...
+	"github.com/harvester/harvester-network-controller/pkg/config"
+	"github.com/harvester/harvester-network-controller/pkg/utils"
+)
+
+const (
+	ControllerName = "harvester-network-manager-nad-controller"
+
+	jobContainerName      = "network-helper"
+	jobServiceAccountName = "harvester-network-helper"
+	JobEnvName            = "NADNETWORKS"
```
guangbochen, line 30: \s\NADNETWORKS\NAD_NETWORKS

pkg/controller/manager/nad/controller.go
```diff
@@ -0,0 +1,334 @@
...
+		return nad, nil
+	}
+
+	if networkConf.Mode != utils.Auto {
+		return nad, nil
+	}
+	// create or update job to get layer 3 network information automatically
+	if err := h.createOrUpdateJob(nad); err != nil {
+		return nil, err
+	}
```
guangbochen, line 117: need to sync the configured status from the job pod to network CR.

pkg/utils/nadconf.go
yaocw2020: I wonder what is the DNS for here. DHCP IP? Do you mean the IP of the DHCP server?

pkg/controller/manager/nad/controller.go
yaocw2020: We don't need to sync the status because if the job runs successfully, the connectivity will be set as `unknown` in the annotation of nad CR, otherwise, it will be empty.

pkg/utils/nadconf.go
guangbochen: Yes the IP of the DHCP server, the DNS is redundant so please ignore here.

yaocw2020: I will try to sent DHCPINFORM message instead of DHCPDISCOVERY to get the DHCP server options.

cmd/network-helper/main.go
```diff
@@ -0,0 +1,86 @@
...
+			Name:   "kubeconfig, k",
+			EnvVar: "KUBECONFIG",
+			Value:  "",
+			Usage:  "Kubernetes config files, e.g. $HOME/.kube/config",
+		},
+		cli.StringFlag{
+			Name:   "master, m",
+			EnvVar: "MASTERURL",
+			Value:  "",
+			Usage:  "Kubernetes cluster master URL.",
```
gitlawr, line 33: Let's name it "server" like the kubeconfig way

pkg/network/vlan/vlan.go
```diff
@@ -172,18 +175,57 @@ func (v *Vlan) stopMonitor() {
...
+		return fmt.Errorf("add bridge vlan %d failed, error: %w", id, err)
+	}
+
+	if cidr == "" {
+		return nil
+	}
+
+	if err := iface.EnsureRouteViaGateway(cidr); err != nil {
+		return fmt.Errorf("ensure %s to route via gateway failed, error: %w", cidr, err)
+	}
```
gitlawr, line 196: We delete routes on vlan setup. Here we add routes again. Does it help if we don't clean up routes at the begining?

yaocw2020: No, they are different routes.

cmd/network-helper/main.go
yaocw2020: It's called masterurl in the function `BuildConfigFromFlags` of client-go. It would be better to keep the same. 
``` go
// BuildConfigFromFlags is a helper function that builds configs from a master
// url or a kubeconfig filepath. These are passed in as command line flags for cluster
// components. Warnings should reflect this usage. If neither masterUrl or kubeconfigPath
// are passed in we fallback to inClusterConfig. If inClusterConfig fails, we fallback
// to the default config.
func BuildConfigFromFlags(masterUrl, kubeconfigPath string) (*restclient.Config, error) {
	if kubeconfigPath == "" && masterUrl == "" {
		klog.Warningf("Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.")
		kubeconfig, err := restclient.InClusterConfig()
		if err == nil {
			return kubeconfig, nil
		}
		klog.Warning("error creating inClusterConfig, falling back to default config: ", err)
	}
	return NewNonInteractiveDeferredLoadingClientConfig(
		&ClientConfigLoadingRules{ExplicitPath: kubeconfigPath},
		&ConfigOverrides{ClusterInfo: clientcmdapi.Cluster{Server: masterUrl}}).ClientConfig()
}
```

gitlawr (approved):
LGTM

Jellyfrog:
Since this adds support for setting dhcp helper, will it solve this issue? https://github.com/harvester/harvester/issues/1557

.drone.yml
```diff
@@ -57,6 +57,27 @@ steps:
       event:
         - push
 
+  - name: docker-publish-master
```
guangbochen, line 60: use a sperate name of the network-controller's one

.drone.yml
```diff
@@ -76,6 +97,25 @@ steps:
       event:
         - tag
 
+  - name: docker-publish
```
guangbochen, line 100: ditto

guangbochen:
> Since this adds support for setting dhcp helper, will it solve this issue? [harvester/harvester#1557](https://github.com/harvester/harvester/issues/1557)

probably not, since u are using the VMWare ESXi, probably u need to set the vSwitch VLAN ID to `4095` to allow all packets to pass through.
image

package/Dockerfile
```diff
@@ -1,4 +1,10 @@
-FROM alpine:3.9
-RUN apk add ebtables iptables
+FROM alpine:3.12 as network-controller
+# ebtables v2.0.10-4
+# iptables v1.6.2
+RUN echo -e 'http://dl-cdn.alpinelinux.org/alpine/v3.9/main' > /etc/apk/repositories && apk update && apk add ebtables iptables
```
guangbochen, line 4: can u elaborate on the comment of why need to tag specific version here also per quote from the CI:
```
Use the `--no-cache` switch to avoid the need to use `--update` and remove `/var/cache/apk/*` when done installing packages (DL3019)
```

manifests/daemonset.yaml
```diff
@@ -14,6 +14,7 @@ spec:
       name: harvester-network-controller
   template:
     metadata:
+      namespace: harvester-system
```
guangbochen, line 17: the namespace should not be required.

guangbochen (approved):
LGTM, @yaocw2020  can u submit a chart PR of those manifest file changes, thanks.

yaocw2020:
nad example(auto mode): 
```
apiVersion: k8s.cni.cncf.io/v1
kind: NetworkAttachmentDefinition
metadata:
  annotations:
    network.harvesterhci.io/layer3-network: '{"mode":"auto","cidr":"172.16.178.0/24","gateway":"172.16.178.1","connectivity":"true"}'
  creationTimestamp: "2021-11-09T15:38:07Z"
  finalizers:
  - wrangler.cattle.io/harvester-network-manager-nad-controller
  - wrangler.cattle.io/harvester-network-nad-controller
  generation: 1
  labels:
    network.harvesterhci.io/vlan-id: "178"
    networks.harvesterhci.io/type: L2VlanNetwork
  name: vlan178
  namespace: default
  resourceVersion: "5744176"
  uid: 993ae289-d3b1-400f-9bf6-a5f7eafbab19
spec:
  config: '{"cniVersion":"0.3.1","name":"vlan178","type":"bridge","bridge":"harvester-br0","promiscMode":true,"vlan":178,"ipam":{}}'
```

PR #49, opened 2021-11-24
Modify CI

yaocw2020:


guangbochen:
prefer to follow the same patter like https://github.com/harvester/harvester/blob/master/package/Dockerfile.webhook

yaocw2020:
> prefer to follow the same patter like https://github.com/harvester/harvester/blob/master/package/Dockerfile.webhook

Updated

guangbochen (approved):
LGTM

PR #50, opened 2021-11-24
No need to wait down signal after setting no master

yaocw2020:
The network interface will not be down after setting nomaster from the bridge in the new harvester-os.

Related issue: 
https://github.com/harvester/harvester/issues/1464
https://github.com/harvester/harvester/issues/1529


gitlawr (approved):
LGTM

guangbochen (approved):

PR #51, opened 2021-11-27
Set nad vlan-id label

yaocw2020:
Related issue: https://github.com/harvester/harvester/issues/1476

### DIFF
--- a/pkg/controller/agent/nad/controller.go
+++ b/pkg/controller/agent/nad/controller.go
@@ -2,18 +2,16 @@ package nad
 
 import (
 	"context"
-	"encoding/json"
 	"errors"
+	"fmt"
 	"strconv"
 
-	hn "github.com/harvester/harvester/pkg/api/network"
 	ctlcniv1 "github.com/harvester/harvester/pkg/generated/controllers/k8s.cni.cncf.io/v1"
 	nadv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
 	"github.com/vishvananda/netlink"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/klog"
 
-	networkv1 "github.com/harvester/harvester-network-controller/pkg/apis/network.harvesterhci.io/v1beta1"
 	"github.com/harvester/harvester-network-controller/pkg/config"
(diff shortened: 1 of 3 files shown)


pkg/controller/manager/nad/controller.go
```diff
@@ -120,6 +131,23 @@ func (h Handler) OnChange(key string, nad *cniv1.NetworkAttachmentDefinition) (*
 	return nad, nil
 }
 
+func (h Handler) ensureLabels(nad *cniv1.NetworkAttachmentDefinition) error {
```
gitlawr, line 134: what’s the purpose of this function?

yaocw2020: To add the label `network.harvesterhci.io/vlan-id"

pkg/controller/manager/nad/controller.go
```diff
@@ -232,6 +260,8 @@ func constructJob(cur *batchv1.Job, namespace, image, dhcpServerAddr string, nad
 			ServiceAccountName: jobServiceAccountName,
 		},
 	}
+	backoffLimit := int32(2)
+	job.Spec.BackoffLimit = &backoffLimit
```
guangbochen, line 264: should we consider cleaning up the pod if the job is succeeded to avoid data overwhelming?

pkg/controller/agent/nad/controller.go
```diff
@@ -70,13 +67,17 @@ func (h Handler) OnChange(key string, nad *nadv1.NetworkAttachmentDefinition) (*
...
 	// TODO delete previous vlan id when update nad
 
+	vlanStr, ok := nad.Labels[utils.KeyVlanLabel]
+	if !ok {
+		return nad, nil
+	}
+	vlanID, err := strconv.Atoi(vlanStr)
+	if err != nil {
+		return nil, fmt.Errorf("invalid vlan %s", vlanStr)
+	}
```
guangbochen, line 79: seems a duplicate code in many places, can we refactor it?

yaocw2020: Both the VLAN id with string format and integer format are required. I think it's fine here.

pkg/controller/manager/nad/controller.go
yaocw2020: I suggest reserving the pods for debugging if there is something wrong.

gitlawr (approved):
LGTM

guangbochen (approved):

PR #53, opened 2021-12-14
Ignore default PVID

yaocw2020:
Related issue: https://github.com/harvester/harvester/issues/1669

The command to configure PVID is different from other VIDs. Assume we
want to configure PVID as 1 and add VID 10 for harvester-mgmt, the commands are as followings.
- Configure PVID 1
  `bridge vlan add vid 1 dev harvester-mgmt pvid untagged master`
- Add VID 10
  `bridge vlan add vid 10 dev harvester-mgmt master`
The default PVID is 1, so we ignore it.

### DIFF
--- a/pkg/network/iface/link.go
+++ b/pkg/network/iface/link.go
@@ -24,6 +24,8 @@ const (
 
 	tableFilter  = "filter"
 	chainForward = "FORWARD"
+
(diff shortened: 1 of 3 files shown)


gitlawr (approved):
LGTM

guangbochen (approved):
LGTM

PR #57, opened 2022-03-03
Migrate base image to BCI image

yaocw2020:
We won't install ebtables in the new image.
The harvester-br0 only inherits the IP address and routes of the
harvester-mgmt network interface. Meanwhile, the IP address of
the harvester-mgmt could not be changed. Therefore, it's not necessary
to allow the DHCP package to pass to the VLAN interface anymore.

Related issue: https://github.com/harvester/harvester/issues/1868

### DIFF
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -1,7 +1,6 @@
-FROM alpine:3.12
-# ebtables v2.0.10-4
-# iptables v1.6.2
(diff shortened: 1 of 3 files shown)


tjjh89017 (approved):
LGTM

guangbochen (approved):

PR #58, opened 2022-06-06
Bump golang to 1.18

weihanglo:
related: harvester/harvester#2267

This is necessary because we want to upgrade k8s dependency to 1.23 in harvester, and k8s 1.23 client-go requires golang 1.17+ to build. However, network-controller is lag behind with golang 1.16.


weihanglo:
FYI: I'll submit a follow-up PR to upgrade k8s client library for this project.

guangbochen (approved):

yaocw2020 (approved):
LGTM

PR #59, opened 2022-08-23
Vlan enhancement

yaocw2020:
Related issue: https://github.com/harvester/harvester/issues/2236

image


pkg/webhook/nad/mutator.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+var _ types.Mutator = &nadMutator{}
+
+func NewNadMutator() *nadMutator {
+	return &nadMutator{}
+}
+
+func (n *nadMutator) Create(request *types.Request, newObj runtime.Object) (types.Patch, error) {
+	netAttachDef := newObj.(*cniv1.NetworkAttachmentDefinition)
```
w13915984028, line 28: check type converting, or assume it is always `cniv1.NetworkAttachmentDefinition` ?

pkg/webhook/vlanconfig/mutator.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/harvester/harvester-network-controller/pkg/utils"
+)
+
+type vlanConfigMutator struct {
+	types.DefaultMutator
+
+	nodeCache ctlcorev1.NodeCache
+}
+
+var _ types.Mutator = &vlanConfigMutator{}
```
w13915984028, line 22: Question: what's the usage of this unamed global variable?

pkg/webhook/vlanconfig/mutator.go
```diff
@@ -0,0 +1,125 @@
...
+	selector, err := utils.NewSelector(vc.Spec.NodeSelector)
+	if err != nil {
+		return nil, err
+	}
+	nodes, err := v.nodeCache.List(selector)
+	if err != nil {
+		return nil, err
+	}
+
+	matchedNodes := make([]string, len(nodes))
```
w13915984028, line 90: check `len(nodes) == 0 ` ?

pkg/webhook/nad/mutator.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	netconf := &utils.NetConf{}
+	if err := json.Unmarshal([]byte(netAttachDef.Spec.Config), netconf); err != nil {
+		return nil, fmt.Errorf(createErr, netAttachDef.Namespace, netAttachDef.Name, err)
+	}
+	labels[utils.KeyVlanLabel] = strconv.Itoa(netconf.Vlan)
+
+	lenOfBrName := len(netconf.BrName)
+	if lenOfBrName < 3 {
```
w13915984028, line 42: `3` looks to be `len(iface.BridgeSuffix)`, it's better to also define a constant in iface, replaces the occurances in mutator, validator

pkg/controller/agent/vlanconfig/controller.go
```diff
@@ -0,0 +1,402 @@
...
+	if setupErr != nil {
+		goto updateStatus
+	}
+	// set up VLAN
+	v = vlan.NewVlan(vc.Spec.ClusterNetwork, vids)
+	setupErr = v.Setup(uplink)
+
+updateStatus:
+	// Update status and still return setup error if not nil
+	if err := h.updateStatus(vc, v, setupErr); err != nil {
```
w13915984028, line 148: `setupErr` should be also formatted into return error here, without it, it's difficult to identify it's from `goto` or normal run.

`return fmt.Errorf("update status into vlanstatus %s", h.statusName(vc.Name))`

pkg/controller/agent/vlanconfig/controller.go
```diff
@@ -0,0 +1,402 @@
...
+			teardownErr = nil
+		}
+		goto updateStatus
+	}
+	if teardownErr = v.Teardown(); teardownErr != nil {
+		goto updateStatus
+	}
+
+updateStatus:
+	if err := h.removeNodeLabel(vc); err != nil {
```
w13915984028, line 179: similar,
`teardownErr` is the first root cause, it should be formatted into return error

meanwhile, `log` those import information ?

pkg/network/iface/link.go
```diff
@@ -124,15 +101,15 @@ func (l *Link) clearMacVlan() error {
 }
 
 func (l *Link) SetMaster(br *Bridge, vids []uint16) error {
-	if l.link.Attrs().MasterIndex == br.bridge.Index {
+	if l.Attrs().MasterIndex == br.Index {
 		return nil
 	}
```
w13915984028, line 107: `SetMaster` seems should exclude adding `vids`

It's `vlanconfig`'s task to :
```
  vlanconfig--->l.setMaster(as bridge)
  vlanconfig----loop---->..l.AddBridgeVlan(vid)
```

pkg/network/vlan/vlan.go
```diff
@@ -2,190 +2,130 @@ package vlan
...
-		if err := v.bridge.SyncIPv4Addr(l); err != nil {
-			return err
-		}
-		if err := v.bridge.ToLink().AddRoutes(l); err != nil {
-			return err
-		}
+		return fmt.Errorf("ensure bridge %s failed, error: %w", v.bridge.Name, err)
 	}
 
 	// set master
```
w13915984028, line 83: vlan to:
```
  set uplink's master as bridge
  (loop) set all vids to uplink's vlan filter
```  

link's `SetMaster` should not combine `set master` and `set vlan filter`

pkg/network/vlan/vlan.go
```diff
@@ -2,190 +2,130 @@ package vlan
...
-	if v.nic == nil {
-		return fmt.Errorf("physical nic vlan network")
+	if ok, _ := v.findVid(vid); ok {
+		return nil
 	}
 
-	if err := v.nic.AddBridgeVlan(uint16(id)); err != nil {
-		return fmt.Errorf("add bridge vlan %d failed, error: %w", id, err)
+	if err := v.uplink.AddBridgeVlan(vid); err != nil {
+		return fmt.Errorf("add bridge vlanconfig %d failed, error: %w", vid, err)
```
w13915984028, line 126: In such a path, considering reconciler
```
v.findVid (return when existing)

v.uplink.AddBridgeVlan
v.vids = append(v.vids, vid)
iface.EnsureRouteViaGateway(cidr)
```

1. `v.vids = append(v.vids, vid)` seems should be put in last,  before return nil finally

2. what's the return result of `v.uplink.AddBridgeVlan` for duplicated vlan ?  in case it returns error, then add an vlan id existing check in error case (inside `AddBridgeVlan`)  is roubust.

pkg/webhook/vlanconfig/mutator.go
yaocw2020: Indicate that `&vlanConfigMutator{}` must implement the interface `types.Mutator`.

pkg/controller/manager/vlanconfig/controller.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	if len(vcs) != 1 {
+		return nil
+	}
+	return h.cnClient.Delete(name, &metav1.DeleteOptions{})
+}
+
+func (h Handler) initialize() error {
+	// It's not allowed to use the local cache to get the cluster network in the register period
```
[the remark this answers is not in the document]
yaocw2020: It'd be better to retry in a goroutine to avoid crashing the program.

pkg/controller/manager/nad/controller.go
```diff
@@ -120,59 +138,31 @@ func (h Handler) OnChange(key string, nad *cniv1.NetworkAttachmentDefinition) (*
...
-		return nad, nil
+		return nil
 	}
 
 	if networkConf.Mode != utils.Auto {
-		return nad, nil
+		return nil
 	}
 	// create or update job to get layer 3 network information automatically
 	if err := h.createOrUpdateJob(nad, networkConf.ServerIPAddr); err != nil {
```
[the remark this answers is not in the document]
yaocw2020: Good catch. I ignore to add the required node affinity to make the pod of job to be scheduled to the expected nodes.

pkg/network/vlan/vlan.go
```diff
@@ -194,89 +134,53 @@ func (v *Vlan) AddLocalArea(id int, cidr string) error {
...
-		return fmt.Errorf("remove bridge vlan %d failed, error: %w", id, err)
+	ok, index := v.findVid(vid)
+	if !ok {
+		return nil
 	}
 
+	if err := v.uplink.DelBridgeVlan(vid); err != nil {
+		return fmt.Errorf("remove bridge vlanconfig %d failed, error: %w", vid, err)
+	}
+	v.vids = append(v.vids[:index], v.vids[index+1:]...)
```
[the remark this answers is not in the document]
yaocw2020: It's not out of range and a normal way to delete item from slice.

pkg/controller/agent/vlanconfig/controller.go
```diff
@@ -0,0 +1,402 @@
...
+			linkAttrs.TxQLen = vc.Spec.Uplink.LinkAttrs.TxQLen
+		}
+		if vc.Spec.Uplink.LinkAttrs.HardwareAddr != nil {
+			linkAttrs.HardwareAddr = vc.Spec.Uplink.LinkAttrs.HardwareAddr
+		}
+	}
+	// Note: do not use &netlink.Bond{}
+	bond := netlink.NewLinkBond(linkAttrs)
+	// set bonding mode
+	mode := netlink.BOND_MODE_BALANCE_TLB
```
tjjh89017, line 210: plz use active-backup as default
TLB has lots of problems with bridges

bk201: @tjjh89017 Could you briefly add the issue TLB might cause here?

tjjh89017: TLB will modify the src mac for most packets for egress.
That will cause the switch will not detect some macs for a long time unless an ARP response is sent.
If MAC entry in the switch is expired, it will cause too much BUM traffic.

pkg/network/vlan/vlan.go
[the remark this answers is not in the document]
tjjh89017: I'm not sure of the reason not to allow all VLAN directly.

pkg/webhook/nad/mutator.go
yaocw2020: Yes, it should be always cniv1.NetworkAttachmentDefinition. If not and something happen out of expectation, it's better to fix it in the dependency package.

pkg/network/vlan/vlan.go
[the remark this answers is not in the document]
yaocw2020: @tjjh89017 It sounds like a good idea.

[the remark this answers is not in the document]
yaocw2020: Because the netlink package could add only one VLAN ID at one time, allow all VLAN directly means that the program will execute thousands of command when creating or updating a bond. It sounds not elegant.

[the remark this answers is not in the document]
yaocw2020: Because the netlink package could add only one VLAN ID at one time, allow all VLAN directly means that the program will execute thousands of commands when creating or updating a bond. It sounds not elegant.

pkg/network/iface/bond.go
```diff
@@ -81,16 +82,21 @@ func (b *Bond) ensureBondSlaves() error {
...
+			if err := netlink.LinkSetDown(l); err != nil {
+				return fmt.Errorf("set slave %s down failed, error: %w", slave, err)
+			}
 			if err := netlink.LinkSetBondSlave(l, b.Bond); err != nil {
 				return fmt.Errorf("add slave %s to bond %s failed, error: %w", slave, b.Name, err)
 			}
 		}
 
-		if l.Attrs().OperState != netlink.OperUp {
+		if l.Attrs().Flags != net.FlagUp {
```
tjjh89017, line 94: It seems Flags is multiple conditions in one variable.
We should use bitwise to test.

```suggestion
		if l.Attrs().Flags & net.FlagUp == 0 {
```

yaocw2020: Good catch.

pkg/network/vlan/vlan.go
```diff
@@ -114,58 +119,58 @@ func (v *Vlan) Teardown() error {
...
 		return nil
 	}
 
-	if err := v.uplink.AddBridgeVlan(vid); err != nil {
-		return fmt.Errorf("add bridge vlanconfig %d failed, error: %w", vid, err)
+	if err := v.uplink.AddBridgeVlan(la.Vid); err != nil {
+		return fmt.Errorf("add bridge vlanconfig %d failed, error: %w", la.Vid, err)
 	}
-	v.vids = append(v.vids, vid)
+	v.localAreas = append(v.localAreas, la)
```
w13915984028, line 133: put `append vid` behind `iface.EnsureRouteViaGateway`, for reconciler to proper retry

question: the return result of `v.uplink.AddBridgeVlan(vid)` when vid is existing?  considering reconciler's retry

w13915984028 (commented):
The new commit 2fb4eace999 solves most of the review comments. Thanks.

The last 2 are listed, please double check.

pkg/network/vlan/vlan.go
yaocw2020: It won't return any error if the vid is existing.

w13915984028 (approved):
LGTM, thanks.

tjjh89017 (approved):
LGTM

pkg/apis/network.harvesterhci.io/v1beta1/clusternetwork.go
```diff
@@ -1,24 +1,13 @@
...
 // +genclient
 // +genclient:nonNamespaced
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 // +kubebuilder:resource:shortName=cn;cns,scope=Cluster
-// +kubebuilder:printcolumn:name="DESCRIPTION",type=string,JSONPath=`.spec.description`
-// +kubebuilder:printcolumn:name="ENABLE",type=boolean,JSONPath=`.enable`
 
 type ClusterNetwork struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
```
bk201, line 12: Still not quite sure if it's OK to remove fields in CRD.
I manually apply the updated CRD manifest and the only clusternetwork `vlan` seems to survive because most of the fields are empty.

bk201: @yaocw2020 is the old `vlan` CR supposed to be removed in the upgrade path?

yaocw2020: Maybe, I am considering removing them directly or just tagging them as deprecated.

pkg/controller/agent/vlanconfig/controller.go
```diff
@@ -0,0 +1,428 @@
+package vlanconfig
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"github.com/harvester/harvester-network-controller/pkg/apis/network.harvesterhci.io"
```
guangbochen, line 8: format fix.

guangbochen (commented):
For some feedbacks, we can tentatively leave it for rc2 if it is less important, thanks.

bk201 (approved):

cmd/webhook/main.go
```diff
@@ -0,0 +1,169 @@
...
+	ctlcorev1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
+	"github.com/rancher/wrangler/pkg/kubeconfig"
+	"github.com/rancher/wrangler/pkg/leader"
+	"github.com/rancher/wrangler/pkg/signals"
+	"github.com/rancher/wrangler/pkg/start"
+	"github.com/sirupsen/logrus"
+	"github.com/urfave/cli"
+	"github.com/yaocw2020/webhook/pkg/config"
+	"github.com/yaocw2020/webhook/pkg/server"
+	"github.com/yaocw2020/webhook/pkg/types"
```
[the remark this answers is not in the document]
yaocw2020: OK

.drone.yml
```diff
@@ -78,6 +78,27 @@ steps:
...
+  - name: docker-publish-master-webhook
+    image: plugins/docker
+    settings:
+      build_args:
+        - ARCH=amd64
+        - VERSION=${DRONE_BRANCH}-${DRONE_COMMIT_SHA:0:8}-head
+      dockerfile: package/Dockerfile.webhook
+      password:
+        from_secret: docker_password
+      repo: "rancher/harvester-network-webhook"
```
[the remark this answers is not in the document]
yaocw2020: No.

pkg/apis/network.harvesterhci.io/v1beta1/clusternetwork.go
```diff
@@ -1,24 +1,13 @@
...
-	// +optional
```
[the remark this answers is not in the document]
yaocw2020: I'll handle in the RC2 version.

pkg/apis/network.harvesterhci.io/v1beta1/nodenetwork.go
```diff
@@ -1,127 +0,0 @@
-package v1beta1
```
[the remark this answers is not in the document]
yaocw2020: ditto

pkg/controller/manager/common/mgmt.go
```diff
@@ -0,0 +1,3 @@
+package common
+
+const ManagementClusterNetworkName = "mgmt"
```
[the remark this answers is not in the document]
yaocw2020: As a prefix, harvester-mgmt is too long, more than 12 characters, while hvst-mgmt isn't suitable to be the name of clusterNetwork.

pkg/apis/network.harvesterhci.io/v1beta1/clusternetwork.go
yaocw2020: I'll consider it in the RC2 version.

guangbochen (approved):

PR #60, opened 2022-09-26
Reject vlanconfigs overlap with others and add clusternetwork status and webhook

yaocw2020:
Solution two issues in one PR for easy testing.

Related issue: 
https://github.com/harvester/harvester/issues/2828

Solution: 
Reject the vlanconfigs that overlap with others directly in the webhook validator.

Related issue: 
https://github.com/harvester/harvester/issues/2841

Solution: 
Add clusternetwork status and webhook
- Add clusternetwork status. Set the status as ready if there is any vlanstatu under the clusternetwork，vice versa.
- Add clusternetwork validator to forbid to delete clusternetwork under which there are vlanconfigs existing.

Additional fixed problem:
Don't need to add/delete vids to/from mgmt uplink

Related issue:
https://github.com/harvester/harvester/issues/2734
Solution:
 Reserve nodenetwork CRD and tag it deprecated

Dependency PR:
https://github.com/harvester/charts/pull/115


pkg/webhook/vlanconfig/validator.go
```diff
@@ -41,6 +41,11 @@ var _ types.Validator = &vlanConfigValidator{}
 func (v *vlanConfigValidator) Create(_ *types.Request, newObj runtime.Object) error {
 	vc := newObj.(*networkv1.VlanConfig)
 
+	if vc.Spec.ClusterNetwork == utils.ManagementClusterNetworkName {
+		return fmt.Errorf(createErr, vc.Name, fmt.Errorf("cluster network could not be %s",
+			utils.ManagementClusterNetworkName))
+	}
```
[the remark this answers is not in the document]
yaocw2020: It's not allowed to add vlanconfigs under mgmt clusternetwork. Users can select mgmt when creating NAD.

pkg/apis/network.harvesterhci.io/v1beta1/vlanconfig.go
```diff
@@ -22,10 +23,12 @@ type VlanConfig struct {
...
 type VlanConfigSpec struct {
 	// +optional
-	Description    string            `json:"description,omitempty"`
-	ClusterNetwork string            `json:"clusterNetwork"`
-	NodeSelector   map[string]string `json:"nodeSelector,omitempty"`
-	Uplink         Uplink            `json:"uplink"`
+	Description    string `json:"description,omitempty"`
+	ClusterNetwork string `json:"clusterNetwork"`
+	// optional
+	Paused       bool              `json:"paused,omitempty"`
```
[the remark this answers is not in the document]
yaocw2020: Sounds good.

pkg/webhook/vlanconfig/mutator.go
```diff
@@ -78,24 +80,31 @@ func getCnLabelPatch(v *networkv1.VlanConfig) types.Patch {
...
+			return nil, err
+		} else if err == nil && vs.Status.VlanConfig != vc.Name {
+			// The vlanconfig is found means a vlanconfig with the same clusternetwork
+			// has been taken effect on this node. We have to pause the new vlanconfig.
+			pausedNodes = append(pausedNodes, node.Name)
+		}
+		matchedNodes = append(matchedNodes, node.Name)
 	}
+
+	return matchedNodesToPatch(vc, matchedNodes, pausedNodes)
```
guangbochen, line 104: instead of patching the spec would it be easier to add the check on the validator and reject to create directly?

pkg/controller/agent/vlanconfig/controller.go
```diff
@@ -98,35 +100,54 @@ func (h Handler) OnChange(key string, vc *networkv1.VlanConfig) (*networkv1.Vlan
...
-			return true, node.Labels[utils.KeyVlanConfigLabel], nil
+			return true, nil
 		}
 	}
 
-	return false, "", nil
+	return false, nil
+}
+
+func (h Handler) inEffect(vc *networkv1.VlanConfig) (bool, error) {
```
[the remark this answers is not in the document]
yaocw2020: This word `ineffect` means the exact opposite of that phrase `in effect`. Let me find a more proper word.

pkg/webhook/vlanconfig/mutator.go
yaocw2020: Yes. It's logically more direct.

pkg/webhook/vlanconfig/validator.go
```diff
@@ -90,7 +116,38 @@ func (v *vlanConfigValidator) Resource() types.Resource {
...
+
+	pausedNodes := make([]string, 0, len(matchedNodes))
+	for _, node := range matchedNodes {
+		vsName := utils.Name("", vc.Spec.ClusterNetwork, node)
+		if vs, err := v.vsCache.Get(vsName); err != nil && !apierrors.IsNotFound(err) {
+			return err
+		} else if err == nil && vs.Status.VlanConfig != vc.Name {
+			// The vlanconfig is found means a vlanconfig with the same clusternetwork  has been taken effect on this node.
+			pausedNodes = append(pausedNodes, node)
```
guangbochen, line 144: you should be able to return the error message in here directly.

yaocw2020: I think it's better to return all the overlap nodes in the error message.

pkg/controller/agent/vlanconfig/controller.go
```diff
@@ -98,35 +100,54 @@ func (h Handler) OnChange(key string, vc *networkv1.VlanConfig) (*networkv1.Vlan
 }
 
 func (h Handler) OnRemove(key string, vc *networkv1.VlanConfig) (*networkv1.VlanConfig, error) {
+	if vc == nil {
+		return nil, nil
+	}
+
+	if isInEffect, err := h.isApplied(vc); err != nil {
```
w13915984028, line 107: change `isInEffect` to `isApplied`,  straightforward

pkg/webhook/vlanconfig/validator.go
tjjh89017: I will also consider returning all nodes in one error message here.

tjjh89017 (approved):
LGTM

pkg/controller/manager/vlanconfig/controller.go
```diff
@@ -56,62 +57,115 @@ func (h Handler) OnChange(key string, vc *networkv1.VlanConfig) (*networkv1.Vlan
...
 		return err
 	}
 
-	if len(vcs) != 1 {
+	if networkv1.Ready.IsTrue(cn.Status) {
 		return nil
 	}
-	return h.cnClient.Delete(name, &metav1.DeleteOptions{})
+	cnCopy := cn.DeepCopy()
+	networkv1.Ready.True(&cnCopy.Status)
```
w13915984028, line 118: when network's status is same as the desired, avoid update

w13915984028 (commented):
Mostly are done, please check remaining 2, thanks.

pkg/controller/manager/vlanconfig/controller.go
yaocw2020: We check it in line 114.

pkg/controller/agent/vlanconfig/controller.go
```diff
@@ -73,22 +73,24 @@ func (h Handler) OnChange(key string, vc *networkv1.VlanConfig) (*networkv1.Vlan
...
+
+	if !isMatched {
+		if isApplied {
+			if err := h.removeVLAN(vc); err != nil {
+				return nil, err
+			}
+		}
 		return vc, nil
 	}
+
```
w13915984028, line 93: Last question:

`OnChange` may trigger `setupVLAN` mutli times, how to avoid duplicated set, can `isApplied` be helpful?

yaocw2020: Yes, it doesn't matter to call `setupVLAN` multiple times. `isApplied` is used to confirm whether the vlanconfig is applied into the node, but it's powerless to check whether is required to setup VLAN again.

w13915984028 (approved):
LGTM, thanks.

guangbochen (approved):

PR #62, opened 2022-10-11
Remove vlan.localAreas

yaocw2020:
Related issue: https://github.com/harvester/harvester/issues/2903

### DIFF
--- a/pkg/controller/agent/vlanconfig/controller.go
+++ b/pkg/controller/agent/vlanconfig/controller.go
@@ -166,12 +166,15 @@ func (h Handler) setupVLAN(vc *networkv1.VlanConfig) error {
 		goto updateStatus
 	}
 	// set up VLAN
-	v = vlan.NewVlan(vc.Spec.ClusterNetwork, localAreas)
-	setupErr = v.Setup(uplink)
+	v = vlan.NewVlan(vc.Spec.ClusterNetwork)
+	if setupErr = v.Setup(uplink); setupErr != nil {
+		goto updateStatus
+	}
+	setupErr = h.AddLocalAreas(v, localAreas)
 
 updateStatus:
 	// Update status and still return setup error if not nil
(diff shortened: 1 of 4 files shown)


w13915984028 (approved):
LGTM, thanks.

pkg/controller/agent/vlanconfig/controller.go
```diff
@@ -166,12 +166,15 @@ func (h Handler) setupVLAN(vc *networkv1.VlanConfig) error {
 		goto updateStatus
 	}
 	// set up VLAN
-	v = vlan.NewVlan(vc.Spec.ClusterNetwork, localAreas)
-	setupErr = v.Setup(uplink)
+	v = vlan.NewVlan(vc.Spec.ClusterNetwork)
+	if setupErr = v.Setup(uplink); setupErr != nil {
+		goto updateStatus
```
guangbochen, line 171: a `goto` method is preferred when the normal method does not apply to you, in here I would suggest us to refactor to call a normal method in the future version.

pkg/controller/agent/vlanconfig/controller.go
```diff
@@ -166,12 +166,15 @@ func (h Handler) setupVLAN(vc *networkv1.VlanConfig) error {
...
+	}
+	setupErr = h.AddLocalAreas(v, localAreas)
```
guangbochen, line 173: whey do we ignore the setupErr here?

yaocw2020: We don't ignore it. The follow codes is to handle the error when updating the status.

pkg/controller/agent/vlanconfig/controller.go
yaocw2020: OK

guangbochen (approved):
LGTM, thanks.

PR #63, opened 2022-10-17
Add untagged network

yaocw2020:
Related issue: https://github.com/harvester/harvester/issues/1618

### DIFF
--- a/pkg/controller/manager/clusternetwork/controller.go
+++ b/pkg/controller/manager/clusternetwork/controller.go
@@ -7,6 +7,7 @@ import (
 	"github.com/cenk/backoff"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog/v2"
 
 	networkv1 "github.com/harvester/harvester-network-controller/pkg/apis/network.harvesterhci.io/v1beta1"
 	"github.com/harvester/harvester-network-controller/pkg/config"
@@ -111,9 +112,14 @@ func (h Handler) deleteLinkMonitor(name string) error {
 func (h Handler) initialize() error {
 	if err := backoff.Retry(func() error {
 		if err := h.initializeClusterNetwork(); err != nil {
+			klog.V(5).Info(err)
 			return err
 		}
-		return h.initializeLinkMonitor()
+		if err := h.initializeLinkMonitor(); err != nil {
+			klog.V(5).Info(err)
+			return err
+		}
+		return nil
 	}, backoff.NewExponentialBackOff()); err != nil {
 		return err
 	}
--- a/pkg/controller/manager/nad/controller.go
+++ b/pkg/controller/manager/nad/controller.go
@@ -96,6 +96,10 @@ func (h Handler) OnChange(key string, nad *cniv1.NetworkAttachmentDefinition) (*
 		return nil, nil
 	}
 
+	if utils.IsEmptyNAD(nad) {
+		return nad, nil
+	}
+
 	klog.Infof("nad configuration %s has been changed: %s", nad.Name, nad.Spec.Config)
 
 	if err := h.EnsureJob2GetLayer3NetworkInfo(nad); err != nil {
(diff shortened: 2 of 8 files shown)


pkg/controller/manager/clusternetwork/controller.go
```diff
@@ -137,6 +153,26 @@ func (h Handler) initializeClusterNetwork() error {
 	return nil
 }
 
+func (h Handler) initializeUntaggedNetwork() error {
+	untaggedNetworkConfig := fmt.Sprintf(`{"cniVersion":"0.3.1","name":"%s","type":"bridge","bridge":"%s-br","promiscMode":true,"ipam":{}}`,
```
w13915984028, line 157: suggest to use a const to represent the cni version, and format it with that const

pkg/controller/manager/nad/controller.go
```diff
@@ -96,6 +96,11 @@ func (h Handler) OnChange(key string, nad *cniv1.NetworkAttachmentDefinition) (*
 		return nil, nil
 	}
 
+	if nad.Spec.Config == "" || nad.Labels == nil || nad.Labels[utils.KeyVlanLabel] == "" ||
```
w13915984028, line 99: suggest to encap the nad default/empty check into one function, and reuse in `OnRemove`

w13915984028 (approved):
LGTM, please check those 2 suggestions, thanks.

pkg/controller/manager/clusternetwork/controller.go
tjjh89017, line 157: I will suggest to gen the bridge name before the config string
If we change the bridge name rule. we could modify the naming rule.
For example, change the posfix of bridge name.
and we don't need to change here to match the new rule

tjjh89017 (approved):
LGTM

pkg/controller/manager/clusternetwork/controller.go
```diff
@@ -16,24 +19,30 @@ import (
 )
 
 const (
-	controllerName = "harvester-network-manager-cn-controller"
-	nicMonitorName = "nic"
+	controllerName        = "harvester-network-manager-cn-controller"
+	nicMonitorName        = "nic"
+	publicNamespace       = "harvester-public"
+	untaggedNetworkName   = "untagged-network"
+	networkConfigTemplate = `{"cniVersion":"0.3.1","name":"%s","type":"bridge","bridge":"%s","promiscMode":true,"ipam":{}}`
```
w13915984028, line 26: networkConfigTemplate = `{"cniVersion":"0.3.1","name":"%s","type":"bridge","bridge":"%s","promiscMode":true,"ipam":{}}`
->
networkConfigTemplate = '{"cniVersion":"0.3.1","name":"%s","type":"bridge","bridge":"%s","promiscMode":true,"ipam":{}}'

pkg/controller/manager/clusternetwork/controller.go
```diff
@@ -137,6 +155,26 @@ func (h Handler) initializeClusterNetwork() error {
 	return nil
 }
 
+func (h Handler) initializeUntaggedNetwork() error {
+	bridgeName := iface.GenerateName(utils.ManagementClusterNetworkName, iface.BridgeSuffix)
+	untaggedNetworkConfig := fmt.Sprintf(networkConfigTemplate, untaggedNetworkName, bridgeName)
+	untaggedNetwork := &cniv1.NetworkAttachmentDefinition{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: publicNamespace,
```
guangbochen, line 163: any reasons why it should be limited to `harvester-public` ns only and need a default config?

yaocw2020: Because all the users can access the `harvester-public` ns and the VMs across different namespaces can use the NAD in the `harvester-public` ns.

guangbochen: I would prefer to leave the users/admins to create on-demand, exposing untagged-network on the public ns is not a good practice for security and RBAC concerns.

yaocw2020: Agree, let me delete the untagged network initialization and communicate with UI team.

pkg/controller/manager/clusternetwork/controller.go
```diff
@@ -21,19 +23,22 @@ const (
 )
 
 type Handler struct {
-	lmClient ctlnetworkv1.LinkMonitorClient
-	lmCache  ctlnetworkv1.LinkMonitorCache
-	cnClient ctlnetworkv1.ClusterNetworkClient
+	lmClient  ctlnetworkv1.LinkMonitorClient
+	lmCache   ctlnetworkv1.LinkMonitorCache
+	cnClient  ctlnetworkv1.ClusterNetworkClient
+	nadClient ctlcniv1.NetworkAttachmentDefinitionClient
```
guangbochen, line 29: seems this can be removed.

yaocw2020: updated.

guangbochen (approved):
LGTM, thanks.

PR #66, opened 2022-11-18
Fix two issues related with the vlanconfig

yaocw2020:
Related issue:  https://github.com/harvester/harvester/issues/3173
Solution: Ensure labels of vlanconfig in the controller rather than webhook

Related issue: https://github.com/harvester/harvester/issues/3166
Solution:  Keep NICs up after deleting the vlanconfig


pkg/controller/manager/nad/controller.go
```diff
@@ -125,6 +135,41 @@ func (h Handler) OnRemove(key string, nad *cniv1.NetworkAttachmentDefinition) (*
...
+	if netconf.Vlan != 0 {
+		labels[utils.KeyNetworkType] = string(l2VlanNetwork)
+		labels[utils.KeyVlanLabel] = strconv.Itoa(netconf.Vlan)
+	} else {
+		labels[utils.KeyNetworkType] = string(untaggedNetwork)
+	}
+
+	lenOfBrName := len(netconf.BrName)
+	if lenOfBrName < len(iface.BridgeSuffix) {
+		return fmt.Errorf("the length of brName must be more than %d", len(iface.BridgeSuffix))
```
w13915984028, line 162: `more than` or `>=` ?

pkg/network/iface/bond.go
```diff
@@ -110,12 +110,23 @@ func (b *Bond) ensureBondSlaves() error {
...
+		return err
+	}
+
+	if err := netlink.LinkDel(b); err != nil {
 		return err
 	}
 
-	return netlink.LinkDel(b)
+	for _, slave := range slaves {
+		if err := netlink.LinkSetUp(slave); err != nil {
```
w13915984028, line 124: Question:
If some slave is `down` before removing bond, will it cause failure here?

w13915984028 (commented):
few questions, please doube check, thanks.

pkg/network/iface/bond.go
yaocw2020: No, the bond could be deleted even though all slaves are down.

pkg/webhook/nad/validator.go
```diff
@@ -48,8 +49,9 @@ func (n *nadValidator) Create(_ *types.Request, newObj runtime.Object) error {
 		return fmt.Errorf(createErr, netAttachDef.Namespace, netAttachDef.Name, err)
 	}
 
+	// The VLAN value of untagged network will be empty or number 0.
```
guangbochen, line 52: I think we should limit the untagged network to empty only.

pkg/controller/manager/nad/controller.go
```diff
@@ -125,6 +135,43 @@ func (h Handler) OnRemove(key string, nad *cniv1.NetworkAttachmentDefinition) (*
...
+
+	netconf := &utils.NetConf{}
+	if err := json.Unmarshal([]byte(nad.Spec.Config), netconf); err != nil {
+		return fmt.Errorf("unmarshal config of nad %s/%s failed, error: %w", nad.Namespace, nad.Name, err)
+	}
+
+	lenOfBrName := len(netconf.BrName)
+	if lenOfBrName <= len(iface.BridgeSuffix) {
+		return fmt.Errorf("the length of brName must be more than %d", len(iface.BridgeSuffix))
+	}
```
guangbochen, line 156: prefer not to check the bridgeSuffix via controller loop, if needed, move it to the webhook validator.

pkg/controller/manager/nad/controller.go
```diff
@@ -125,6 +135,43 @@ func (h Handler) OnRemove(key string, nad *cniv1.NetworkAttachmentDefinition) (*
...
+
+	labels := nad.Labels
+	if labels == nil {
+		labels = make(map[string]string)
+	}
+
+	netconf := &utils.NetConf{}
+	if err := json.Unmarshal([]byte(nad.Spec.Config), netconf); err != nil {
+		return fmt.Errorf("unmarshal config of nad %s/%s failed, error: %w", nad.Namespace, nad.Name, err)
```
guangbochen, line 150: the error message looks overlapped with #106, we can return the error directly here.

pkg/webhook/nad/validator.go
yaocw2020: The value of vlan field will be 0 after JSON unmarshaling even though the vlan field is empty.

guangbochen (approved):
LGTM and verified that: 
1. both bridge & bond networks can be reset successfully, and it will not set the physical NIC to a down status.
2. is able to re-config the VLAN network and the VM is able to request an IP successfully with the same config.

FrankYang0529 (approved):
LGTM. Thanks.

w13915984028 (approved):
LGTM, thanks.

PR #68, opened 2023-01-30
GetLocalArea function is not available for untagged network

yaocw2020:
Related issue: https://github.com/harvester/harvester/issues/3337


w13915984028 (approved):
LGTM, thanks.

guangbochen (approved):
LGTM, regarding the issue, can we please help to share a proper workaround to mitigate the issue? Thanks.

PR #80, opened 2023-04-18
Avoid bond from recreating periodically

yaocw2020:
**Problem:**

The bond will recreate every 10 hour and cause network flapping. 

**Solution:**

- Set default value in the CRD
- Ignore the default value when judging whether there is a parameter change

**Related Issue:** https://github.com/harvester/harvester/issues/3744

**Dependency:** https://github.com/harvester/charts/pull/155

**Test plan:**

- Create a new cluster network and create the related vlanconfig. The default value of miimon and TxQLen is -1. -1 means to keep the original value which may be system default value or set manually before.
- Delete harvester network controller pod, watch `dmesg -T`, the bond will not be recreated.

### DIFF
--- a/manifests/crds/network.harvesterhci.io_vlanconfigs.yaml
+++ b/manifests/crds/network.harvesterhci.io_vlanconfigs.yaml
@@ -57,11 +57,14 @@ spec:
               uplink:
                 properties:
                   bondOptions:
-                    description: 'TODO: reference: https://www.kernel.org/doc/Documentation/networking/bonding.txt'
+                    description: 'reference: https://www.kernel.org/doc/Documentation/networking/bonding.txt'
                     properties:
                       miimon:
+                        default: -1
+                        minimum: -1
                         type: integer
                       mode:
+                        default: active-backup
                         enum:
                         - balance-rr
                         - active-backup
@@ -80,8 +83,11 @@ spec:
                         format: byte
                         type: string
                       mtu:
+                        minimum: 0
                         type: integer
                       txQLen:
+                        default: -1
+                        minimum: -1
                         type: integer
                     type: object
                   nics:
--- a/pkg/network/iface/bond.go
+++ b/pkg/network/iface/bond.go
@@ -166,10 +166,28 @@ func getSlaves(index int) ([]netlink.Link, error) {
 }
 
 func compareBond(b1, b2 *netlink.Bond) bool {
-	if b1.Name == b2.Name && b1.MTU == b2.MTU && b1.Mode == b2.Mode &&
-		b1.HardwareAddr.String() == b2.HardwareAddr.String() && b1.TxQLen == b2.TxQLen {
-		return true
+	if b1.Name != b2.Name {
+		return false
 	}
 
-	return false
+	if b1.Mode != b2.Mode {
+		return false
+	}
+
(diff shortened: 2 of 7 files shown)


pkg/network/iface/bond.go
```diff
@@ -165,11 +165,34 @@ func getSlaves(index int) ([]netlink.Link, error) {
 	return links, nil
 }
 
-func compareBond(b1, b2 *netlink.Bond) bool {
-	if b1.Name == b2.Name && b1.MTU == b2.MTU && b1.Mode == b2.Mode &&
-		b1.HardwareAddr.String() == b2.HardwareAddr.String() && b1.TxQLen == b2.TxQLen {
-		return true
+func compareBond(old, new *netlink.Bond) bool {
```
futuretea, line 168: It would be nice to have a unit test for this function

yaocw2020: The reason why we need to ignore the default value depend on the package [vishvananda/netlink](https://github.com/vishvananda/netlink). The function is very easy which is the reason why there isn't a unit test.

yaocw2020: Adding more annotations to explain why it's necessary to ignore the default value would be more helpful.

pkg/network/iface/bond.go
```diff
@@ -165,11 +165,36 @@ func getSlaves(index int) ([]netlink.Link, error) {
...
+		return false
+	}
+
+	// skip if mtu is omitted
+	if new.MTU != 0 && old.MTU != new.MTU {
+		return false
+	}
+
+	// skip if hardware addr is omitted
+	if new.HardwareAddr.String() != "" && old.HardwareAddr.String() != new.HardwareAddr.String() {
```
futuretea, line 183: Question:
If the new is "" and the old is not, is returning false what we expect?
Or is the new value the default, but the old value is not the default, and returning false is what we expect?

yaocw2020: - The compare function is just non-exported function and the new bond will never be nil in the current implementation.
- Yes, once the new value is the default, skip it and don't recreate the bond.

futuretea (approved):
LGTM, Thanks!
no problem was found during the test according to the test plan

manifests/crds/network.harvesterhci.io_vlanconfigs.yaml
```diff
@@ -57,11 +57,14 @@ spec:
               uplink:
                 properties:
                   bondOptions:
-                    description: 'TODO: reference: https://www.kernel.org/doc/Documentation/networking/bonding.txt'
+                    description: 'reference: https://www.kernel.org/doc/Documentation/networking/bonding.txt'
                     properties:
                       miimon:
+                        default: -1
```
futuretea, line 63: Question:
why the default value of miimon is -1, default value in https://www.kernel.org/doc/Documentation/networking/bonding.txt is 0

futuretea: `keep the original value` should be `keep the existing value`?

starbops (approved):
Tested with shortened resync period. LGTM, thanks for the fix!

bk201 (approved):
lgtm

PR #83, opened 2023-05-23
Disable iptables bridge forwarding on initialization

yaocw2020:
**Problem:**

In the Harvester cluster whose management network has a VLAN ID, the VM will be unable to access the host port or node port with the host IP where the VM is running.

**Solution:**

Disable net.bridge.bridge-nf-call-iptables to avoid iptables rules affecting bridge forwarding on vlan.init().

We could not disable net.bridge.bridge-nf-call-iptables in the harvester-installer because RKE2 will enable it after harvester-installer.

**Related Issue:** https://github.com/harvester/harvester/issues/3960

**Test plan:**

- Spin up a Harvester whose management network has a VLAN ID
- Create a VM with VLAN network whose VLAN ID is same with the management network.
- Curl nodeIP:443 in the VM.

### DIFF
--- a/pkg/network/iface/bridge.go
+++ b/pkg/network/iface/bridge.go
@@ -31,10 +31,6 @@ func NewBridge(name string) *Bridge {
 // Ensure bridge
 // set promiscuous mod default
 func (br *Bridge) Ensure() error {
-	if err := disableBridgeNF(); err != nil {
-		return fmt.Errorf("disable net.bridge.bridge-nf-call-iptables failed, error: %w", err)
-	}
-
 	if err := netlink.LinkAdd(br); err != nil && err != syscall.EEXIST {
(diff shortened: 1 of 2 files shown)


iosifnicolae2:
Isn't disabling  `net.bridge.bridge-nf-call-iptables` a security risk?

starbops (approved):
LGTM. It works fine with the current Harvester design/implementation. However, I'm unsure of its impact on future network policy-relevant features with either NeuVector or Harvester's own solution (like the one mentioned in the SecurityGroups HEP).

mingshuoqiu (approved):
Agree to @starbops. The issue #3960 also need to disable it.

bk201:
@mergifyio backport v0.5.x

PR #85, opened 2023-06-08
chore: Remove deprecated CRD nodenetworks

yaocw2020:
**Problem:**

We kept the deprecated CRD nodenetworks for upgrading from Harvester v1.0.2 to v1.1.0 before. Because it's not used anymore from Harvester v1.2.0, we remove it.

**Solution:**

**Related Issue:**

**Test plan:**


guangbochen (approved):
LGTM, thanks.

futuretea (approved):
LGTM, thanks.

PR #104, opened 2024-07-31
webhook/mutator: filter out the witness node

Vicente-Cheng:
**Problem:**
The matched-node returns the whole nodes even if we have one witness node.
The witness node did not run any workload, so we should filter out this.

**Solution:**
we do not allow any workload on the witness node, so we should filter out the witness node when creating the vlan config.

**Related Issue:**
https://github.com/harvester/harvester/issues/5325#issuecomment-2236668074

**Test plan:**
Make sure the annotation of vlanconfig response should looks like below (harvester-node-1 is the witness node)
```
annotations: {network.harvesterhci.io/matched-nodes: "["harvester-node-0","harvester-node-2"]"}
```

### DIFF
--- a/pkg/utils/labels.go
+++ b/pkg/utils/labels.go
@@ -20,4 +20,6 @@ const (
 
 	ValueTrue  = "true"
 	ValueFalse = "false"
+
+	HarvesterWitnessNodeLabelKey = "node-role.harvesterhci.io/witness"
 )
--- a/pkg/webhook/vlanconfig/mutator.go
+++ b/pkg/webhook/vlanconfig/mutator.go
@@ -10,6 +10,7 @@ import (
 	admissionregv1 "k8s.io/api/admissionregistration/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/selection"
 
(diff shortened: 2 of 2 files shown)


Vicente-Cheng:
cc @a110605, backend PR is here.

starbops (approved):
LGTM, thank you!

mingshuoqiu (approved):
LGTM. Thanks

a110605:
Hi @Vicente-Cheng , I test with latest [harvester-network-controller](https://hub.docker.com/layers/rancher/harvester-network-controller/master-head/images/sha256-c53ad762fdaf0a89eab0b49f88f481006cec6aeeee4564888d9f2c21d737ffa0?context=explore). The config via select all nodes filter out witness when first created, but wait for 3 ~ 5 mins the witness node is auto added in annotation. 

Do we have any mechanism to update annotation in backend ? 
Reproduce Steps:
1. create network config via select all nodes
2. check the config YAML, the witness node is not in matched-nodes annotation
3. Wait for 3 ~ 5 mins and refresh the page.
4. The witness node is added in annoation

Vicente-Cheng:
@mergifyio backport v0.4.x

Vicente-Cheng:
@mergifyio backport v0.4.x

PR #105, opened 2024-08-05
controller/node: skip update vlanconfig with witness node

Vicente-Cheng:
**Problem:**
Now, the `matched-nodes` will filter out the witness node with creation.
However, the controller has a chance to update the witness node to the `matched-nodes` on the controller part.

**Solution:**
We need to skip to update the `matched-nodes` with the witness node change.

**Related Issue:**
https://github.com/harvester/network-controller-harvester/pull/104#issuecomment-2268030300

**Test plan:**
Make sure the `matched-nodes` annotation of vlanconfig does not contain any witness node.

### DIFF
--- a/pkg/controller/manager/node/controller.go
+++ b/pkg/controller/manager/node/controller.go
@@ -61,6 +61,11 @@ func (h Handler) OnChange(_ string, node *corev1.Node) (*corev1.Node, error) {
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


mingshuoqiu (approved):
LGTM

Vicente-Cheng:
@mergifyio backport v0.4.x

PR #111, opened 2024-08-09
Add validation for max length of cluster network name

rrajendran17:
**Problem:**
when creating a cluster network using GUI, user specifies a name for cluster network and it is validated for maximum length of 12 at the UI code,but when cluster network with name > 12 chars is created using API, the cluster network is created successfully.

cluster network webhook validator is missing the max length check for cluster network name.
Later it is validated when vlan config is applied on the same cluster network.
But the creation of cluster network should validate it.

**Solution:**
Added create function in cluster network validator to validate for maximum length of cluster network name.

**Related Issue:**

harvester/harvester#3826

**Test plan:**
Manually tested on a harvester node with the changes.

### DIFF
--- a/pkg/webhook/clusternetwork/validator.go
+++ b/pkg/webhook/clusternetwork/validator.go
@@ -10,10 +10,12 @@ import (
 
 	networkv1 "github.com/harvester/harvester-network-controller/pkg/apis/network.harvesterhci.io/v1beta1"
 	ctlnetworkv1 "github.com/harvester/harvester-network-controller/pkg/generated/controllers/network.harvesterhci.io/v1beta1"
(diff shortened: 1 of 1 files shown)


starbops (approved):
LGTM, thank you!

mingshuoqiu (approved):
LGTM. Thank you.

lanfon72:
@bk201 does this need backport to old versions?

bk201:
yes, we need a backport. I [updated some comments about the workflow ](https://github.com/harvester/harvester/issues/3826#issuecomment-2288070407)and I moved the issue back to review first because the change didn't land in v1.4-head either.

PR #113, opened 2024-08-15
3826 - backport Add validation for cluster network name length

rrajendran17:
**Problem:**
cluster network name length is not validated for maximum length when created using API

**Solution:**
Add validation for cluster network max length in webhook validator of cluster network

**Related Issue:**

**Test plan:**
tested manually on harvester node


mingshuoqiu (approved):
LGTM. Thanks

starbops (approved):
LGTM, thank you!

PR #115, opened 2024-09-09
2995 check if storage network attached before cluster network and vlan config config deletion

rrajendran17:
**Problem:**
1.User can delete the cluster network when storage network NAD is still attached to it.
2.User can delete the vlan config when storage network NAD is still attached to it.

The above breaks the networking functionality of storage networks without user being notified.

**Solution:**
Do not delete the cluster network or vlan config if storage network still exists for the cluster.

**Related Issue:**

**Test plan:**
Manually validated in a harvester node.

Reference bug report:
https://github.com/harvester/tests/issues/1311

Negative: Delete cluster network when already set on storage network

    Navigate to Networks -> Cluster Networks/Configs, create Cluster Network named vlan2
    Create Network Config vlan2-config for all nodes
    Navigate to Advanced -> Settings, edit storage-network
    Select Enable then select vlan2 as cluster network, fill in VLAN ID and IP Range
    Navigate to Networks -> Cluster Networks/Configs, delete Cluster Network vlan2
    Warning or error message should displayed

Negative: Delete vlan config when already set on storage network

    Navigate to Networks -> Cluster Networks/Configs, create Cluster Network named vlan2
    Create Network Config vlan2-config for all nodes
    Navigate to Advanced -> Settings, edit storage-network
    Select Enable then select vlan2 as cluster network, fill in VLAN ID and IP Range
    Navigate to Networks -> Cluster Networks/Configs, delete network config vlan2-config
    Warning or error message should displayed

### DIFF
--- a/pkg/webhook/clusternetwork/validator.go
+++ b/pkg/webhook/clusternetwork/validator.go
@@ -3,7 +3,6 @@ package clusternetwork
 import (
 	"fmt"
 
-	ctlcniv1 "github.com/harvester/harvester/pkg/generated/controllers/k8s.cni.cncf.io/v1"
 	"github.com/harvester/webhook/pkg/server/admission"
 	admissionregv1 "k8s.io/api/admissionregistration/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -16,23 +15,20 @@ import (
 )
 
 const (
-	createErr                           = "could not create cluster network %s because %w"
-	deleteErr                           = "could not delete cluster network %s because %w"
-	StorageNetworkNetAttachDefNamespace = "harvester-system"
+	createErr = "could not create cluster network %s because %w"
+	deleteErr = "could not delete cluster network %s because %w"
 )
 
 type CnValidator struct {
 	admission.DefaultValidator
-	vcCache  ctlnetworkv1.VlanConfigCache
-	nadCache ctlcniv1.NetworkAttachmentDefinitionCache
+	vcCache ctlnetworkv1.VlanConfigCache
 }
 
 var _ admission.Validator = &CnValidator{}
 
-func NewCnValidator(vcCache ctlnetworkv1.VlanConfigCache, nadCache ctlcniv1.NetworkAttachmentDefinitionCache) *CnValidator {
+func NewCnValidator(vcCache ctlnetworkv1.VlanConfigCache) *CnValidator {
 	validator := &CnValidator{
-		vcCache:  vcCache,
-		nadCache: nadCache,
+		vcCache: vcCache,
 	}
 	return validator
 }
@@ -55,17 +51,6 @@ func (c *CnValidator) Delete(_ *admission.Request, oldObj runtime.Object) error 
 
 	if cn.Name == utils.ManagementClusterNetworkName {
 		return fmt.Errorf(deleteErr, cn.Name, fmt.Errorf("it's not allowed"))
-	}
-
-	nads, err := c.nadCache.List(StorageNetworkNetAttachDefNamespace, labels.Set(map[string]string{
-		utils.KeyClusterNetworkLabel: cn.Name,
-	}).AsSelector())
-	if err != nil {
-		return fmt.Errorf(deleteErr, cn.Name, err)
-	}
-
-	if len(nads) > 0 {
-		return fmt.Errorf(deleteErr, cn.Name, fmt.Errorf("storage network is still attached"))
 	}
 
 	vcs, err := c.vcCache.List(labels.Set{
--- a/pkg/webhook/clusternetwork/validator.go
+++ b/pkg/webhook/clusternetwork/validator.go
@@ -3,6 +3,7 @@ package clusternetwork
 import (
 	"fmt"
 
+	ctlcniv1 "github.com/harvester/harvester/pkg/generated/controllers/k8s.cni.cncf.io/v1"
 	"github.com/harvester/webhook/pkg/server/admission"
 	admissionregv1 "k8s.io/api/admissionregistration/v1"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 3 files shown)


mingshuoqiu (approved):
LGTM. Thanks

pkg/webhook/clusternetwork/validator.go
```diff
@@ -53,6 +57,17 @@ func (c *CnValidator) Delete(_ *admission.Request, oldObj runtime.Object) error
 		return fmt.Errorf(deleteErr, cn.Name, fmt.Errorf("it's not allowed"))
 	}
 
+	nads, err := c.nadCache.List(StorageNetworkNetAttachDefNamespace, labels.Set(map[string]string{
```
w13915984028, line 60: This may be redundant.

The following `vlanconfig` checks if any vc on this cluster is still existing, then block it. The `storagenetwork` also relys on one `vlanconfig`. Checking `vlanconfig` seems to be enough here.

On the other hand, below function on Harvester needs to check `VlanConfig` is existing when utilize this `vlanconfig` to carry the storagenetwork.

```
func (v *settingValidator) validateStorageNetworkHelper(value string) error {
```

w13915984028 (commented):
One question, thanks.

pkg/webhook/clusternetwork/validator.go
rrajendran17: @w13915984028 yes makes sense, if storage network is validated for vlan config to be present then validating at cluster network will be redundant.I will remove that check.Thanks.

Regarding storage network validating if vlan config is present, I have another pull request (testing still in progress)
where I am checking if cluster network is ready (cluster network becomes ready only if at least one vlan config attached) before configuring storage networks.
https://github.com/harvester/harvester/pull/6538

pkg/webhook/vlanconfig/validator.go
```diff
@@ -143,6 +144,17 @@ func (v *Validator) Delete(_ *admission.Request, oldObj runtime.Object) error {
...
 
+	nads, err := v.nadCache.List(StorageNetworkNetAttachDefNamespace, labels.Set(map[string]string{
+		utils.KeyClusterNetworkLabel: vc.Spec.ClusterNetwork,
+	}).AsSelector())
+	if err != nil {
+		return fmt.Errorf(deleteErr, vc.Name, err)
+	}
+
+	if len(nads) > 0 {
+		return fmt.Errorf(deleteErr, vc.Name, fmt.Errorf("storage network is still attached"))
```
starbops, line 155: Here we filter NADs only with the CN key. Stating that "storage network is still attached" feels weird. We can't be so sure it's a "storage" network. Technically, Harvester users can create NADs under the `harvester-system` NS. When those VCs get deleted with their associated NADs still existing, the webhook will show such a message, which is probably inaccurate.

rrajendran17: Not sure if there are use cases where users could create NAD via other methods which creates nads in harvester-system namespace.Our harvester by design creating all other non storage network nads only in default namespace.But its good not to assume, we could check the nad's name prefix for "storagenetwork" but to be even stricter, storage network's nad's annotation field is updated specifically to identify its as storage network nad during creation.
So I have used that to identify the nad specifically as "storage network" nad.

starbops (approved):
LGTM, thank you!

pkg/webhook/vlanconfig/validator.go
```diff
@@ -23,9 +23,11 @@ import (
 )
 
 const (
-	createErr = "could not create vlanConfig %s because %w"
-	updateErr = "could not update vlanConfig %s because %w"
-	deleteErr = "could not delete vlanConfig %s because %w"
+	createErr                           = "could not create vlanConfig %s because %w"
+	updateErr                           = "could not update vlanConfig %s because %w"
+	deleteErr                           = "could not delete vlanConfig %s because %w"
+	StorageNetworkNetAttachDefNamespace = "harvester-system"
```
w13915984028, line 29: Please check if possible to refter this namespace definition
```
vendor/github.com/harvester/harvester/pkg/util/constants.go:	HarvesterSystemNamespaceName      = "harvester-system"
```

pkg/webhook/vlanconfig/validator.go
```diff
@@ -23,9 +23,11 @@ import (
...
+	StorageNetworkAnnotation            = "storage-network.settings.harvesterhci.io"
```
w13915984028, line 30: `storage-network` should also define those to consts for common usage, let's update Harvester consts in another PR ,  later change here

```
pkg/controller/master/storagenetwork/storage_network.go:	StorageNetworkAnnotation        = "storage-network.settings.harvesterhci.io"
pkg/controller/master/storagenetwork/storage_network.go:	longhornStorageNetworkName = "storage-network"
```

pkg/webhook/vlanconfig/validator.go
```diff
@@ -143,6 +145,21 @@ func (v *Validator) Delete(_ *admission.Request, oldObj runtime.Object) error {
...
+	nads, err := v.nadCache.List(StorageNetworkNetAttachDefNamespace, labels.Set(map[string]string{
+		utils.KeyClusterNetworkLabel: vc.Spec.ClusterNetwork,
+	}).AsSelector())
+	if err != nil {
+		return fmt.Errorf(deleteErr, vc.Name, err)
+	}
+
+	if len(nads) > 0 {
+		for _, nad := range nads {
+			if nad.Annotations[StorageNetworkAnnotation] == "true" {
```
w13915984028, line 157: to be more secure,  please skip the case where `nad.DeletionTimestamp != nil`

w13915984028 (approved):
LGTM, thanks.

pkg/webhook/vlanconfig/validator.go
rrajendran17: sure added the changes

pkg/webhook/vlanconfig/validator.go
rrajendran17: yes, they are not in common place now

pkg/webhook/vlanconfig/validator.go
rrajendran17: sure added the changes

WebberHuang1118 (approved):
LGTM, thanks for the PR.

bk201:
@mergifyio backport v0.5.x

PR #116, opened 2024-09-11
Sync MTU and check MTU valid values

w13915984028:
**Problem:**

Change of MTU on VlanConfig uplink is not updated to NAD.

**Solution:**

1. Check the input MTU on VlanConfig
2. Update the MTU to labels of ClusterNetwork when change happens
3. Sync NAD's MTU with the lables of ClusterNetwork when change happens

This solution does not change the APIs and UI, instead it propagates MTU from VlanConfig to ClusterNetwork when change happens; and makes sure all the VlanConfigs under the same ClusterNetwork will have the same MTU.

**Related Issue:**
https://github.com/harvester/harvester/issues/4355
https://github.com/harvester/harvester/issues/4752

HEP: https://github.com/harvester/harvester/pull/6385
Doc PR: https://github.com/harvester/docs/pull/640

**Test plan:**


w13915984028:
Replaced by https://github.com/harvester/network-controller-harvester/pull/149

PR #119, opened 2024-09-19
backport: Disable iptables bridge forwarding on initialization

mingshuoqiu:
(cherry picked from commit 652162909d48f81d2759c49538e2ac42ecd75eb9)
Signed-off-by: Chris Chiu <chris.chiu@suse.com>

**Problem:**
In the Harvester cluster whose management network has a VLAN ID, the VM will be unable to access the host port or node port with the host IP where the VM is running.

**Solution:**
Disable net.bridge.bridge-nf-call-iptables to avoid iptables rules affecting bridge forwarding on vlan.init().

We could not disable net.bridge.bridge-nf-call-iptables in the harvester-installer because RKE2 will enable it after harvester-installer.

**Related Issue:**
https://github.com/harvester/harvester/issues/3960

**Test plan:**
* Spin up a Harvester whose management network has a VLAN ID
* Create a VM with VLAN network whose VLAN ID is same with the management network.
* Curl nodeIP:443 in the VM.

### DIFF
--- a/pkg/network/iface/bridge.go
+++ b/pkg/network/iface/bridge.go
@@ -31,10 +31,6 @@ func NewBridge(name string) *Bridge {
 // Ensure bridge
 // set promiscuous mod default
 func (br *Bridge) Ensure() error {
-	if err := disableBridgeNF(); err != nil {
-		return fmt.Errorf("disable net.bridge.bridge-nf-call-iptables failed, error: %w", err)
-	}
-
 	if err := netlink.LinkAdd(br); err != nil && err != syscall.EEXIST {
 		return fmt.Errorf("add iface failed, error: %w, iface: %v", err, br)
 	}
@@ -67,7 +63,7 @@ func (br *Bridge) Ensure() error {
 	return br.Fetch()
 }
 
-func disableBridgeNF() error {
+func DisableBridgeNF() error {
 	return utils.EnsureSysctlValue(bridgeNFCallIptables, "0")
 }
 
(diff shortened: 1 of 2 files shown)


rrajendran17:
@mingshuoqiu I see iptables are disabled by default in the harvester node.Am I missing something ?
renuka-v131-dev:~ # sysctl -a | grep iptable
net.bridge.bridge-nf-call-iptables = 0

mingshuoqiu:
> @mingshuoqiu I see iptables are disabled by default in the harvester node.Am I missing something ? renuka-v131-dev:~ # sysctl -a | grep iptable net.bridge.bridge-nf-call-iptables = 0

It is 1 if no secondary guest network created.

starbops:
Yes, it'll be turned off once there a new ClusterNetwork and VlanConfig created.
https://github.com/harvester/network-controller-harvester/blob/7384dcc0485e280dc2c0f64d7f2949b1fa60fb9d/pkg/network/iface/bridge.go#L34

starbops (approved):
LGTM, thanks.

rrajendran17:
Is this a specific case of mgmt network in custom vlan ? who is the caller of init() ?

bk201 (approved):

mingshuoqiu:
@rrajendran17 I installed Harvester 1.3.2-rc2 on a new machine w/o any config file, the `net.bridge.bridge-nf-call-iptables` is still. We still need this patch to disable it by default.

PR #123, opened 2024-10-16
agent: move the iptables bridge forwarding disable to agent register

mingshuoqiu:
The DisableBridgeNF() placed in init() would cause phantom error message while restart/kill the network-manager pod. Move it to agent register to prevent the unexpected invoke.

Fixes: 652162909d48 ("Disable iptables bridge forwarding on initializatio")

**Problem:**
https://github.com/harvester/harvester/issues/6747

**Solution:**
Move the DisableBridgeNF() to agent register to prevent the unexpected invoke.

**Related Issue:**

**Test plan:**
1. Manually restarting the `harvester-network-controller-manager` deployment or simply killing off the pod
2. check if there's the error message `disable net.bridge.bridge-nf-call-iptables failed, error: open /proc/sys/net/bridge/bridge-nf-call-iptables: read-only file system`
3.

### DIFF
--- a/pkg/controller/agent/vlanconfig/controller.go
+++ b/pkg/controller/agent/vlanconfig/controller.go
@@ -62,6 +62,10 @@ func Register(ctx context.Context, management *config.Management) error {
 		cnCache:    cns.Cache(),
 	}
 
+	if err := handler.initialize(); err != nil {
+		return fmt.Errorf("initialize error: %w", err)
+	}
+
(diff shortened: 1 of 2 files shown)


starbops (approved):
Confirmed that

- Manager pods no longer crash
- Only agent pods set the kernel tunable
- Each node's `net.bridge.bridge-nf-call-iptables` is `0`.

LGTM, thank you.

bk201 (approved):

starbops:
@mergify backport v0.5.x

PR #131, opened 2024-12-04
Use VMCache instead of VMICache to judge if the NAD is in use

mingshuoqiu:
The VMI status will be none if the VM is turned off from the GUI. The VMI information can't not be relied to make sure if particular NAD is attched to any VM or not. Use VMCache instead to prevent from deleting the NAD if the VM is turned off.

**Problem:**
Fail to delete an Off VM if its network was deleted. The VM network can be deleted if the attached VM is off. It will be a problem if user try to turn the VM back to on.

**Solution:**
Do not allow VM network removing if the VM still exists, even it is OFF.

**Related Issue:**
https://github.com/harvester/harvester/issues/6961

**Test plan:**
1. Create a VM network
2. Create a VM and attached to the created VM network
3. Turn the VM from Running to Off.
4. Remove the VM network. Should not be allowed.

### DIFF
--- a/pkg/utils/vmi.go
+++ b/pkg/utils/vmi.go
@@ -14,6 +14,34 @@ type VmiGetter struct {
 	VmiCache ctlkubevirtv1.VirtualMachineInstanceCache
 }
 
+type VMGetter struct {
+	VMCache ctlkubevirtv1.VirtualMachineCache
+}
+
+// WhoUseNad requires adding network indexer to the vmi cache before invoking it
+func (v *VMGetter) VMUseNad(nad *nadv1.NetworkAttachmentDefinition) ([]*kubevirtv1.VirtualMachine, error) {
+	// multus network name can be <networkName> or <namespace>/<networkName>
+	// ref: https://github.com/kubevirt/client-go/blob/148fa0d1c7e83b7a56606a7ca92394ba6768c9ac/api/v1/schema.go#L1436-L1439
+	networkName := fmt.Sprintf("%s/%s", nad.Namespace, nad.Name)
+	vms, err := v.VMCache.GetByIndex(indexeres.VMByNetworkIndex, networkName)
+	if err != nil {
+		return nil, err
+	}
+
+	vmsTmp, err := v.VMCache.GetByIndex(indexeres.VMByNetworkIndex, nad.Name)
+	if err != nil {
+		return nil, err
+	}
+	for _, vm := range vmsTmp {
+		if vm.Namespace != nad.Namespace {
+			continue
+		}
+		vms = append(vms, vm)
+	}
+
+	return vms, nil
+}
+
 // WhoUseNad requires adding network indexer to the vmi cache before invoking it
 func (v *VmiGetter) WhoUseNad(nad *nadv1.NetworkAttachmentDefinition, nodesFilter mapset.Set[string]) ([]*kubevirtv1.VirtualMachineInstance, error) {
 	// multus network name can be <networkName> or <namespace>/<networkName>
--- a/pkg/webhook/nad/validator.go
+++ b/pkg/webhook/nad/validator.go
@@ -25,13 +25,16 @@ const (
 type Validator struct {
 	admission.DefaultValidator
 	vmiCache ctlkubevirtv1.VirtualMachineInstanceCache
+	vmCache  ctlkubevirtv1.VirtualMachineCache
 }
 
 var _ admission.Validator = &Validator{}
 
-func NewNadValidator(vmiCache ctlkubevirtv1.VirtualMachineInstanceCache) *Validator {
+func NewNadValidator(vmiCache ctlkubevirtv1.VirtualMachineInstanceCache,
+	vmCache ctlkubevirtv1.VirtualMachineCache) *Validator {
 	return &Validator{
 		vmiCache: vmiCache,
+		vmCache:  vmCache,
 	}
 }
 
@@ -96,6 +99,10 @@ func (v *Validator) Delete(_ *admission.Request, oldObj runtime.Object) error {
 		return fmt.Errorf(deleteErr, nad.Namespace, nad.Name, err)
 	}
 
+	if err := v.checkVM(nad); err != nil {
+		return fmt.Errorf(deleteErr, nad.Namespace, nad.Name, err)
+	}
+
 	return nil
 }
 
@@ -123,6 +130,24 @@ func (v *Validator) checkNadConfig(bridgeConf *utils.NetConf) error {
 func (v *Validator) checkRoute(config string) error {
 	_, err := utils.NewLayer3NetworkConf(config)
 	return err
+}
+
+func (v *Validator) checkVM(nad *cniv1.NetworkAttachmentDefinition) error {
+	vmGetter := utils.VMGetter{VMCache: v.vmCache}
+	vms, err := vmGetter.VMUseNad(nad)
+	if err != nil {
+		return err
+	}
+
+	if len(vms) > 0 {
+		vmNameList := make([]string, 0, len(vms))
+		for _, vm := range vms {
+			vmNameList = append(vmNameList, vm.Namespace+"/"+vm.Name)
+		}
+		return fmt.Errorf("VM(s) %s still attached which must be removed first", strings.Join(vmNameList, ", "))
+	}
+
+	return nil
 }
 
 func (v *Validator) checkVmi(nad *cniv1.NetworkAttachmentDefinition) error {
--- a/pkg/webhook/vlanconfig/validator.go
+++ b/pkg/webhook/vlanconfig/validator.go
@@ -38,18 +38,21 @@ type Validator struct {
 	vcCache  ctlnetworkv1.VlanConfigCache
 	vsCache  ctlnetworkv1.VlanStatusCache
 	vmiCache ctlkubevirtv1.VirtualMachineInstanceCache
+	vmCache  ctlkubevirtv1.VirtualMachineCache
 }
 
 func NewVlanConfigValidator(
 	nadCache ctlcniv1.NetworkAttachmentDefinitionCache,
 	vcCache ctlnetworkv1.VlanConfigCache,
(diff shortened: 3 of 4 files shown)


rrajendran17:
@mingshuoqiu can you also add steps in Testplan for verifying deletion of vlan config as well with stopped vm.

pkg/utils/vmi.go
```diff
@@ -47,5 +52,36 @@ func (v *VmiGetter) WhoUseNad(nad *nadv1.NetworkAttachmentDefinition, nodesFilte
...
+
+	vms, err := v.VMCache.GetByIndex(indexeres.VMByNetworkIndex, networkName)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, vm := range vms {
+		if vm.Namespace != nad.Namespace {
+			continue
+		}
```
rrajendran17, line 70: is this check needed, as networkName contains both "Namespace + name" and vms are retrieved by matching both.

pkg/webhook/vlanconfig/validator.go
```diff
@@ -137,15 +136,6 @@ func getAffectedNodes(oldCn, newCn string, oldNodes, newNodes mapset.Set[string]
...
-		return fmt.Errorf(deleteErr, vc.Name, err)
-	}
-
-	if err := v.checkVmi(vc, nodes); err != nil {
-		return fmt.Errorf(deleteErr, vc.Name, err)
-	}
-
 	nads, err := v.nadCache.List(util.HarvesterSystemNamespaceName, labels.Set(map[string]string{
 		utils.KeyClusterNetworkLabel: vc.Spec.ClusterNetwork,
 	}).AsSelector())
```
rrajendran17, line 141: The namespace must be "" or metav1.NamespaceAll as we need to fetch all nads from all namespaces.
HarvesterSystemNamespaceName will fetch nads only from harvester-system namespace

rrajendran17:
Thanks @mingshuoqiu for taking care of the comments.
Overall LGTM, except few minor comments.Can you please validate the code changes for nad deletion, vlan config deletion with vms stopped scenarios and add the steps to testplan also.Thanks.

pkg/utils/vmi.go
```diff
@@ -47,5 +52,36 @@ func (v *VmiGetter) WhoUseNad(nad *nadv1.NetworkAttachmentDefinition, nodesFilte
...
+	}
+
+	vmsTmp, err := v.VMCache.GetByIndex(indexeres.VMByNetworkIndex, nad.Name)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, vm := range vmsTmp {
+		if vm.Namespace != nad.Namespace {
+			continue
```
rrajendran17, line 81: I am not sure about this existing check and if its needed, we can still attach a nad from any namespace to a vm in any namespace.This operation is supported currently.And with this check we might miss vms that have attached a nad from another namespace and the nad could be deleted while the vm is still present.

starbops (commented):
I have a general question regarding the fix: Do we still need the VmiGetter? Under what circumstances do we rely on VMIs to check whether an action is allowed or not? Maybe the logic in the `WhoUseNad` for VMI and VM is redundant, and we can completely replace VMI with VM. Please advise. Thank you.

mingshuoqiu:
Need to re-implement the `WhoUseNad` since the ` indexeres.VMByNetworkIndex` has been removed from harvester v1.4. Will replace the `WhoUseNad` by different logic.

mingshuoqiu:
> I have a general question regarding the fix: Do we still need the VmiGetter? Under what circumstances do we rely on VMIs to check whether an action is allowed or not? Maybe the logic in the `WhoUseNad` for VMI and VM is redundant, and we can completely replace VMI with VM. Please advise. Thank you.
 
I tried to replace all VMI handler to VM. But I will get the log from the webhook pod. How should I add the resource to the access list? CRD? @starbops 
```
W0116 09:04:53.736063       1 reflector.go:547] github.com/rancher/lasso/pkg/cache/cache.go:145: failed to list *v1.VirtualMachine: virtualmachines.kubevirt.io is forbidden: User "system:serviceaccount:harvester-system:harvester-network-webhook" cannot list resource "virtualmachines" in API group "kubevirt.io" at the cluster scope
E0116 09:04:53.736170       1 reflector.go:150] github.com/rancher/lasso/pkg/cache/cache.go:145: Failed to watch *v1.VirtualMachine: failed to list *v1.VirtualMachine: virtualmachines.kubevirt.io is forbidden: User "system:serviceaccount:harvester-system:harvester-network-webhook" cannot list resource "virtualmachines" in API group "kubevirt.io" at the cluster scope
```

rrajendran17:
@mingshuoqiu Also please add a step to testcase for NAD update scenario and it should not be allowed on non-running VMs as well.Thanks

w13915984028:
@mingshuoqiu What's the current status & plan of this PR? Will it be merged on v1.5.0? thanks.

The https://github.com/harvester/network-controller-harvester/pull/149 is better to on top of this PR, thus it can have an unified check on NAD, VM, VMI, storagenetwork.

w13915984028:
> harvester-network-webhook...

@mingshuoqiu Need to add vm to the clusterrole

https://github.com/harvester/charts/blob/761090116ddb657e4850308ebb551cf44c67f43f/charts/harvester-network-controller/templates/rbac.yaml#L124

mingshuoqiu:
I'm verifying with the updated rbac. Will update when verification done.

pkg/webhook/nad/validator.go
```diff
@@ -126,18 +126,18 @@ func (v *Validator) checkRoute(config string) error {
 }
 
 func (v *Validator) checkVmi(nad *cniv1.NetworkAttachmentDefinition) error {
```
w13915984028, line 128: This fix, will break the `update nad` feature, as `checkVmi` is also used in `func (v *Validator) Update(_ *admission.Request, oldObj, newObj runtime.Object)`.

When user change nad, all VMIs need to be stopped. If it is storagenetwork nad, user must clear storagenetwork/set-to-another storagenetwork nad first.

When user delete nad, all VMs need to be changed to remove the network referrences. If it is storagenetwork nad, user must clear storagenetwork/set-to-another storagenetwork nad first.

A new function `checkVM` is needed for nad deletion webhook.

Other changes need to reviewed similarly.

w13915984028: @rrajendran17  Change the `Vlan,HairpinMode,PromiscMode and some others`  of the NAD params is reasonable, stop & start VM is okay;  it is not a good idea to force user to delete the VM/VM's related interface.


```
type NetConf struct {
	cniv1.NetConf
	BrName       string `json:"bridge"`
	IsGW         bool   `json:"isGateway"`
	IsDefaultGW  bool   `json:"isDefaultGateway"`
	ForceAddress bool   `json:"forceAddress"`
	IPMasq       bool   `json:"ipMasq"`
	MTU          int    `json:"mtu"`
	HairpinMode  bool   `json:"hairpinMode"`
	PromiscMode  bool   `json:"promiscMode"`
	Vlan         int    `json:"vlan"`
}
```


change `BrName` (`clusternetwork`) is rare or should be denied.

w13915984028:
More about the scenarios:

In current stage, the most import param of NAD is VID. 

(1) Currently, NAD is allowed to be changed, as long as all attached VMIs are stopped.

If we force the VM should also remove the referrences to NAD before NAD change, it is a break change. The broken scenarios are like:

(a) User migrates a group of VMs from VID 100 to VID 200 duo to network re-plan; or change the L3 IP subnets for this VID.

(b) User creates a dedicated NAD for network testing, and creates a couple of VMs attaching to this NAD. User plans to open a new VID 300, stop this group of VMs, change NAD, start the VMs, then test the new VID.  User can use this groups of VMs to test any existing VIDs at any time.

(c) ...

(2) Change the `BrName (clusternetwork)` of NAD.

This is a bug on Harvester now, it can be done if (1) is met, but, all the related VMs are not reflected. Or I missed something, Harvester has code to care it. Please remind me if you know.

Luckily, we did not receive related bug reports now, if it is an real bug.


We need to:

(i) Enhance (1) via this PR https://github.com/harvester/network-controller-harvester/pull/131, add additional check on VM when NAD is deleted. 

Be very careful, do NOT import break change, VM is the core feature of Harvester.

(ii) Enhance (2) via issue https://github.com/harvester/network-controller-harvester/pull/131, deny the change of `BrName (clusternetwork)`.

---

>But IMO, we should allow such updates on NADs which is not referenced by any of the VMs.

@rrajendran17 There are race cases, create VM & change NAD happen in parallel, it may still cause some issues. 

When a NAD is attached to wrong clusternetwork (very rarely), delete & recreate is not so unaccetable. We may assume this is a corner case, and add this limitation to the document.

If some user really needs this, we can try to support it later.

w13915984028:
FYI:

When deleting a NAD, alone with vmi, vm is also checked. This will be included on PR https://github.com/harvester/network-controller-harvester/pull/149 to cooprate with other functions, I will invite you to review the related parts when it is ready.

https://github.com/harvester/charts/pull/339 is also filed to update rbac.

mingshuoqiu:
@w13915984028 Could you help review again? Thanks

w13915984028:
@mingshuoqiu This depends on the release strategy by @bk201 @starbops 

For issue https://github.com/harvester/harvester/issues/6961, this PR https://github.com/harvester/network-controller-harvester/pull/131 can prevent new `delete vm` operations;  for those existing dangling VMs, the error is still there.

For Harvester v1.5.0, if we want to add a fix, I would recommend https://github.com/harvester/harvester/pull/7743, it fixes the root cause of issue https://github.com/harvester/harvester/issues/6961.  Any exising VMs with dangling NADs can be deleted successfully.

Discussed with @bk201, he prefers to add this PR to v1.5.1/v1.6.0.

For v1.5.1/v1.6.0, check VM when deleting NAD is already a part of https://github.com/harvester/network-controller-harvester/pull/149, and there are more APIs and test codes.

Please @bk201 @starbops take a new look, thanks.

bk201:
let's do it in 1.5.1, thanks!

PR #143, opened 2025-01-09
Update otelhttp and dependencies

mingshuoqiu:
Problem:
Fix the compile error caused by module dependency https://github.com/harvester/network-controller-harvester/pull/135#issuecomment-2560756049

Solution:
Add and update the required packages

Related Issue:

Test plan:


starbops (approved):

innobead:
Outdated.

PR #149, opened 2025-01-30
Sync MTU and check MTU valid values

w13915984028:
**Problem:**

Change of MTU on VlanConfig uplink is not updated to NAD.

**Solution:**

1. Check the input MTU on VlanConfig
2. Update the MTU to labels of ClusterNetwork when change happens
3. Sync NAD's MTU with the lables of ClusterNetwork when change happens

This solution does not change the APIs and UI, instead it propagates MTU from VlanConfig to ClusterNetwork when change happens; and makes sure all the VlanConfigs under the same ClusterNetwork will have the same MTU.

**Related Issue:**
https://github.com/harvester/harvester/issues/4355
https://github.com/harvester/harvester/issues/4752

HEP: https://github.com/harvester/harvester/pull/6385
Doc PR: https://github.com/harvester/docs/pull/640

**Test plan:**

1. MTU propogation
1.1 Create a clusternetwork
1.2 create a vlanconfig, the MTU from vlanconfig is synced to clusternetwork
1.3 create another vlanconfig, if MTU is different, it is denied by webhook
1.4 create nad, the MTU is inherited from clusternetwork/vlanconfig

2. Change the MTU of a clusternetwork
2.1 Stop all VMs attached to a specific clusternetwork
2.2 Delete/Migrate all vlanconfigs until there is a last one; otherwise change MTU from any will be denied per 1.3
2.3 Change the last vlanconfig's MTU, then it is synced to clusternetwork; if there are any NADs attached to this clusternetwork, their MTU is changed automatically
2.4 Add more vlanconfig, each needs to fill the same MTU per 1.3

3. Change the MTU of a mgmt network
3.1 The mgmt network has no vlanconfig instances. You may edit the MTU annotation manually, be sure this really matchs the NICs' MTU, there is no strong check at the moment.
3.2 If there are VMIs or storagenetwork on mgmt network, the change is denied.

More detailed tests about network related enhancements:

https://github.com/harvester/network-controller-harvester/pull/149#issuecomment-2697974336


w13915984028:
@mergify backport v0.5.x v0.6.x

pkg/webhook/vlanconfig/validator.go
```diff
@@ -264,8 +270,8 @@ func (v *Validator) validateMTU(current *networkv1.VlanConfig) error {
 		if vc.Name == current.Name {
```
ibrokethecloud, line 270: would it make sense to change it to just fetch the related cluster object crd and check the MTU from there?

rrajendran17:
Few points
1.Similar to vms,if MTU is updated after storage network configuration, it will be propagated to storage network nad, but LH pods interfaces will not have updated the new MTU.
A webhook check in needed to prevent updating the MTU when storage network is configured. The user has to disable storage network config before MTU change and enable storage network configuration after a new MTU change.
2.In case of storage network, the vlan network under the vlan config also has to be deleted if vlan config has to be deleted for any mismatch in MTU cases.Just a point to be noted while adding test cases.
3.Good to include a manual traffic test after MTU change in the test plan(either between VMs or LH pods with jumbo frames)

pkg/webhook/vlanconfig/validator.go
w13915984028: A few considerations:

1. The current code has assumption that the clusternetwork may be created behind the vlanconfig, for logacy or other reasons
https://github.com/harvester/network-controller-harvester/blob/ce3dd943e183e18bb42a557f02a450eb92cf1b7f/pkg/controller/manager/vlanconfig/controller.go#L86

2. The MTU is synced to clusternetwork as a label, in extreme cases, if user hack it with an invalid value, the check may miss.

w13915984028:
@rrajendran17 Thanks, the storagenetwork related cases need some additional processing, I will add.

pkg/controller/manager/nad/controller.go
```diff
@@ -89,10 +91,63 @@ func Register(ctx context.Context, management *config.Management) error {
 
 	nads.OnChange(ctx, ControllerName, handler.OnChange)
 	nads.OnRemove(ctx, ControllerName, handler.OnRemove)
-
+	cns.OnChange(ctx, ControllerName, handler.OnCNChange)
 	return nil
 }
 
+// Sync cluster network MTU value to all attached NADs
```
rrajendran17, line 98: Why not including this change during cluster network change inside pkg/controller/manager/clusternetwork/controller.go ?

pkg/utils/consts.go
```diff
@@ -0,0 +1,7 @@
+package utils
+
+const (
+	DefaultMTU = 1500
+	MaxMTU     = 9000
+	MinMTU     = 1280 // IPv4 does not define this; IPv6 defines 1280; Harvester adopts it for both
```
rrajendran17, line 6: IIUC, IPv4 defines 576 bytes as minimum MTU.

pkg/webhook/clusternetwork/validator_test.go
```diff
@@ -0,0 +1,284 @@
...
+			newCN: &networkv1.ClusterNetwork{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:        testCnName,
+					Annotations: map[string]string{"test": "test"},
+				},
+			},
+		},
+		{
+			name:      "user can't add the MTU label",
+			returnErr: false,
```
rrajendran17, line 51: since it will return error, returnErr should be true and errKey has error value.

pkg/webhook/clusternetwork/validator_test.go
```diff
@@ -0,0 +1,284 @@
...
+					Uplink: networkv1.Uplink{
+						LinkAttrs: &networkv1.LinkAttrs{
+							MTU: 1500,
+						},
+					},
+				},
+			},
+		},
+		{
+			name:      "ClusterNetwork is allowed to be deleted",
```
rrajendran17, line 229: can you enhance the description ?

pkg/webhook/vlanconfig/validator_test.go
```diff
@@ -0,0 +1,455 @@
...
+
+			err := validator.Delete(nil, tc.currentVC)
+			if tc.returnErr {
+				assert.NotNil(t, err)
+				assert.True(t, strings.Contains(err.Error(), tc.errKey))
+			}
+
+		})
+	}
+}
```
rrajendran17, line 455: If possible, can you also add unit tests for checking running vms and storage network nad configured when updating/deleting vlanconfigs ?

rrajendran17:
I also noticed,currently we do not have min-max MTU check in UI, hope we need to create a GH issue to add those changes to UI as well.

pkg/controller/manager/nad/controller.go
w13915984028: nad controller is the better place to change nad objects

pkg/webhook/vlanconfig/validator_test.go
w13915984028: I am checking PR https://github.com/harvester/network-controller-harvester/pull/131, it has done some checks and is better to be the base of this PR.

pkg/utils/consts.go
rrajendran17: I was checking this from rfc791.
   "All hosts must be prepared
    to accept datagrams of up to 576 octets (whether they arrive whole
    or in fragments).  It is recommended that hosts only send datagrams
    larger than 576 octets if they have assurance that the destination
    is prepared to accept the larger datagrams."

pkg/webhook/nad/validator.go
```diff
@@ -132,14 +168,68 @@ func (v *Validator) checkVmi(nad *cniv1.NetworkAttachmentDefinition) error {
...
+	}
+
+	nt := newNad.Labels[utils.KeyNetworkType]
+	if nt == "" {
+		return fmt.Errorf("nad network type can't be empty")
+	}
+
+	if utils.NetworkType(nt) != utils.L2VlanNetwork && utils.NetworkType(nt) != utils.UntaggedNetwork {
+		return fmt.Errorf("nad network type %s is unknown", nt)
+	}
```
rrajendran17, line 211: Is this check needed ? since we are already checking for empty and there are only two network types defined.

pkg/webhook/nad/validator.go
```diff
@@ -132,14 +168,68 @@ func (v *Validator) checkVmi(nad *cniv1.NetworkAttachmentDefinition) error {
...
+		return fmt.Errorf("old nad does not have label")
+	}
+
+	if newNad.Labels == nil {
+		return fmt.Errorf("new nad does not have label")
+	}
+
+	if oldNad.Labels[utils.KeyClusterNetworkLabel] != newNad.Labels[utils.KeyClusterNetworkLabel] {
+		return fmt.Errorf("nad network can't be changed from %s to %s", oldNad.Labels[utils.KeyClusterNetworkLabel], newNad.Labels[utils.KeyClusterNetworkLabel])
+	}
```
rrajendran17, line 202: As commented in other PR, I still feel that updating a cluster network not attached to any VMs should be allowed.It must be denied only if any VMs referencing it.

A scenario where a NAD is created and not attached to any VMs, and user by mistake attached to a cluster network and wanted to update it to an another cluster network.I think this should at least work as user does not have to delete and recreate the NAD for any update.

rrajendran17:
Thanks @w13915984028 LGTM, posted few questions/clarifications.

pkg/webhook/vlanconfig/validator_test.go
rrajendran17: Thanks

pkg/utils/consts.go
w13915984028: Yes, it is an indirect description, not a directly definition that the MTU should >= 576.

pkg/webhook/nad/validator.go
w13915984028: VxLan may extend, this check may be broken if VxLan add something, then we are notified to work on it.

pkg/webhook/nad/validator.go
w13915984028: Explained here https://github.com/harvester/network-controller-harvester/pull/131#issuecomment-2683153356, more efforts are needed for this corner/rare case.

pkg/controller/manager/nad/controller.go
w13915984028: clusternetwork controller sets the ready condition, but did not touch the details of NAD.

NAD controller decode the NAD config and collaborate with other existing functions/types.

Both of them can do this, looks the NAD controller is the better place.

pkg/webhook/nad/validator.go
w13915984028: Also avoid user to change it to an invalid value

starbops (commented):
Is there a reason to store the MTU value in a label but not an annotation? (I know we saved the VLAN ID value in the label, but I wasn't here, so I didn't have a chance to chime in.)

pkg/utils/consts.go
w13915984028: Could be. Luckily we only need to can change this line. I will update finally when the PR is totally ready.

w13915984028:
With a lot of enforments on ClusterNetwork, VlanConfig, NAD, StorageNetwork, a detailed test plan is listed.

```
test plan:

ClusterNetwork: none-mgmt

Create: 

(1) Create a clusternetwork with MTU annotation: 
  can't create cluster network cntst because label network.harvesterhci.io/uplink-mtu can't be added

Update:

(1) Update a clusternetwork's MTU annotation to another valid value, when there are at least 1 vlanconfig
  can't update cluster network cn3 because clusternetwork has MTU 1430, but the vlanconfigs vc-3-1 has another MTU 1420

(2) Update a clusternetwork's MTU annotation to another valid value, when there is no vlanconfig; is Okay, but when later vlanconfig is added, this value is amended by vlanconfig
  network.harvesterhci.io/uplink-mtu: '1550'

Delete:

(1) A clusternetwork has vlanconfigs
  can't delete cluster network cn4 because vlanconfig(s) [vc-2-1] under this clusternetwork are still existing

(2) A clusternetwork has no vlanconfigs but NADs (NAD is not blocking the deletion of VlanConfig)
  can't delete cluster network cn4 because nads(s) default/nad-2-1 under this clusternetwork are still existing


ClusterNetwork: mgmt

you can't create or delete mgmt clusternetwork

Update

(1) create a NAD, no VM, change mgmt MTU annotation
  OK, mtu is updated to nad: {"cniVersion":"0.3.1","name":"mgmt-nad-1","type":"bridge","bridge":"mgmt-br","promiscMode":true,"vlan":11,"ipam":{},"mtu":1480}

(2) create a NAD, no VM, change mgmt MTU annotation to invalide value
  can't update cluster network mgmt because annotation 1470a value is not an integer, error strconv.Atoi: parsing "1470a": invalid syntax

(3) create a NAD, run a VM, change mgmt MTU annotation to another valid value
  can't update cluster network mgmt because the MTU can't be changed from 1480 to 1490 as following VMs must be stopped at first: default/vm2






note: vlanconfig can only be added to none-mgmt cluster network

VlanConfig: Create

(1) Create the first VlanConfig with an input/default MTU:  
  The MTU is propogate to it's parent clusternetwork's annotation `network.harvesterhci.io/uplink-mtu: '1350'`
  If there are dangling NADs, NADs' MTU are updated

(2) Create a second VlanConfig with a different MTU
  can't create vlanConfig cn-3-2 because the vlanconfig cn-3-2 MTU 1600 is different with another vlanconfig vc-3-1 MTU 0, all vlanconfigs on one clusternetwork need to have same MTU

(3) Create a second VlanConfig with same MTU, OK


VlanConfig: Update

(1) Update: via kubectl, set spec.clusterNetwork to a none-existing CN: (note: From WebUI, the action is `Migrate`, from one clusternetwork to another, the destination CN must be existing)
  can't update vlanConfig vc-2-2 because it refers to a none-existing cluster network cn8 or error clusternetworks.network.harvesterhci.io "cn8" not found;  

(2) Update: via kubectl, set spec.clusterNetwork to `mgmt` CN: 
  can't update vlanConfig vc-2-2 because cluster network can't be mgmt;   

(3.1) Update uplink-mtu to 1440 when there is only 1 vlanconfig; no running VMs
  The updated MTU is propotaged to cluster network annotations: network.harvesterhci.io/uplink-mtu: "1440"
  All NADs under this cluster network are also updated to:
     {"cniVersion":"0.3.1","name":"nad-cn3-3","type":"bridge","bridge":"cn3-br","promiscMode":true,"vlan":300,"ipam":{},"mtu":1440}
     {"cniVersion":"0.3.1","name":"nad-cn3-2","type":"bridge","bridge":"cn3-br","promiscMode":true,"vlan":200,"ipam":{},"mtu":1440}
     {"cniVersion":"0.3.1","name":"nad-cn3-1","type":"bridge","bridge":"cn3-br","promiscMode":true,"vlan":100,"ipam":{},"mtu":1440}

(3.2) Update uplink-mtu to 1440 when there is only 1 vlanconfig; with running VMs
  can't update vlanConfig vc-3-1 because it is blocked by VM(s) default/vm3 which must be stopped at first

(4) Update uplink-mtu to 1450  when there are > 1 vlanconfigs
  can't update vlanConfig vc-3-2 because the vlanconfig vc-3-2 MTU 1450 is different with another vlanconfig vc-3-1 MTU 1440, all vlanconfigs on one clusternetwork need to have same MTU

(5.1) Update selector to expand node (no overlap), OK, no error, the annotation is updated, new node `harv2` is added
   network.harvesterhci.io/matched-nodes: '["harv2","harv41"]'

(5.1) Update selector to reduce node (not affecting VMIs), OK, no error, the annotation is updated, node `harv2` is removed
   network.harvesterhci.io/matched-nodes: '["harv41"]'

(5.2) Update selector to replace current node: running VMIs are affected
  can't update vlanConfig vc-3-1 because it is blocked by VM(s) default/vm3 which must be stopped at first

(6) Migrate to another CN when VMIs are running on related nodes
  can't update vlanConfig vc-3-1 because it is blocked by VM(s) default/vm3 which must be stopped at first  


(7) Update uplink (e.g. nics) when there storage network configured

  can't update vlanConfig cn3-2 because the storage network nad storagenetwork-k8trh is still attached



VlanConfig: Delete

(1) No running VMs, VlanConfig can be deleted, the NADs may be put annotation `network.harvesterhci.io/ready: 'false'` when all VlanConfigs are gone
  This is allowed, as user may re-build the VlanConfig to set other NICs/MTUs...

(2) With running VMs
  can't delete vlanConfig vc-3-1 because it is blocked by VM(s) default/vm3 which must be stopped at first

(3) No running VMs, with storagenetwork
  can't delete vlanConfig cn3-2 because the storage network nad storagenetwork-k8trh is still attached




NAD: Create

(1) create a NAD, refer to a none-existing clusternetwork
  can't create nad default/nad-cn5-1 because clusternetworks.network.harvesterhci.io "cn5" not found

(2) create a NAD, set MTU to a different value than clusternetwork
  the MTU is amended to same value as clusternetwork (via mutator)


NAD: Update

NAD-on none-mgmt network: Update
(1) create a NAD, run a VM, update nad
  can't update nad default/nad-cn3-1 because it's still used by VM(s) default/vm3 which must be stopped at first"]

(2) create a NAD, run a VM, stop VM, update nad: change MTU
  can't update nad default/nad-cn3-3 because nad MTU 1430 does not match cluster network MTU 1420

(3) create a NAD, run a VM, stop VM, update nad: change bridge name
  can't update nad default/nad-cn3-1 because nad bridge name can't be changed from cn3-br to cn4-br"


storagenetwork on none-mgmt network:

(1) Update storagenetwork nad's VID via `kubectl edit NetworkAttachmentDefinition.k8s.cni.cncf.io -n harvester-system`
  can't update nad harvester-system/storagenetwork-k8trh because it is used by storagenetwork



storagenetwork on mgmt network:
(1) edit storagenetwork nad directly via `kubectl edit NetworkAttachmentDefinition.k8s.cni.cncf.io -n harvester-system storagenetwork-c9brt`
  can't update nad harvester-system/storagenetwork-c9brt because nad MTU 1480 does not match cluster network MTU 1470

(2) edit clusternetwork mgmt via `kubectl edit clusternetwork mgmt`
  can't update cluster network mgmt because the MTU can't be changed from 1470 to 1480 as storage network nad storagenetwork-c9brt is still attached




NAD: Delete

NAD-on mgmt network: Delete

(1) create a NAD, run a VM, delete nad
  can't delete nad default/mgmt-nad-1 because it's still used by VM(s) default/vm2 which must be stopped at first

(2) create a NAD, stop a VM, delete nad
  can't delete nad default/mgmt-nad-1 because it's still used by VM(s) default/vm2 which must remove the related networks and interfaces

NAD on none-mgmt network: Delete

(1) create a NAD, run a VM, delete nad
  can't delete nad default/nad-cn3-1 because it's still used by VM(s) default/vm3 which must be stopped at first

(2) create a NAD, stop a VM, delete nad
  can't delete nad default/nad-cn3-1 because it's still used by VM(s) default/vm3 which must remove the related networks and interfaces


storagenetwork on mgmt network:
(1). Disable storagenetwork from Harvester UI, it works. 
  The storagenetwork NAD is deleted by the Harvester controller successfully.


storagenetwork on none-mgmt network:
(1) Update storagenetwork nad via `kk edit NetworkAttachmentDefinition.k8s.cni.cncf.io -n harvester-system`
  can't update nad harvester-system/storagenetwork-k8trh because it is used by storagenetwor
```

w13915984028:
@ibrokethecloud @ihcsim @mschiu77 @starbops @rrajendran17  

This PR is ready for final review, thanks.  This commit https://github.com/harvester/network-controller-harvester/pull/149/commits/ef611ee8acf1a52dc6cc750e8d3775e03091c121 excludes the vendor bump.

With a lot of enhancements on ClusterNetwork, VlanConfig, NAD, StorageNetwork, a detailed test plan is listed https://github.com/harvester/network-controller-harvester/pull/149#issuecomment-2697974336.

cc @bk201

pkg/controller/manager/vlanconfig/controller.go
```diff
@@ -83,16 +83,49 @@ func (h Handler) SetClusterNetworkUnready(_ string, vs *networkv1.VlanStatus) (*
 	return vs, nil
 }
 
-func (h Handler) ensureClusterNetwork(name string) error {
-	if _, err := h.cnCache.Get(name); err != nil && !apierrors.IsNotFound(err) {
+func (h Handler) ensureClusterNetwork(vc *networkv1.VlanConfig) error {
```
ibrokethecloud, line 86: we sync the MTU from the vlanconfig to the clusternetwork. due to additional logic in the validationg webhook we only allow the MTU set on the clusternetwork to be applied to subsequent vlansconfigs. does this mean the only way to change the MTU is remove all vlanconfigs / cluster network and recreate the objects again?

pkg/webhook/clusternetwork/validator.go
```diff
@@ -81,7 +120,102 @@ func (c *CnValidator) Resource() admission.Resource {
...
 		},
 	}
 }
+
+func checkMTUOfNewClusterNetwork(cn *networkv1.ClusterNetwork) error {
+	if cn == nil || cn.Annotations == nil {
+		return nil
+	}
+
+	// for none-mgmt cluster network, this annotation can only be operated by controller
```
ibrokethecloud, line 134: minor typo
```suggestion
	// for non-mgmt cluster network, this annotation can only be operated by controller
```

pkg/utils/vm.go
```diff
@@ -0,0 +1,100 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return generateVMNameList(vms), nil
+}
+
+// Get the vm name list who uses a group of nads,
+// note: duplicated names are removed
+func (v *VMGetter) VMNamesWhoUseNads(nads []*nadv1.NetworkAttachmentDefinition) ([]string, error) {
```
ibrokethecloud, line 73: I could not find this method used anywhere

pkg/utils/vmi.go
```diff
@@ -49,3 +62,61 @@ func (v *VmiGetter) WhoUseNad(nad *nadv1.NetworkAttachmentDefinition, nodesFilte
...
+	if cnt == 0 {
+		return nil, nil
+	}
+
+	if cnt == 1 {
+		return generateVmiNameList(vmis), nil
+	}
+
+	// use mapset to remove duplicated names
+	return mapset.NewSet[string](generateVmiNameList(vmis)...).ToSlice(), nil
```
ibrokethecloud, line 110: is there a specific reason to check `cnt == 1` ? we could just return the mapset here

```suggestion
	return mapset.NewSet[string](generateVmiNameList(vmis)...).ToSlice(), nil
```

w13915984028: When there is only 1 element, surely it has no duplicated names. The `mapset.NewSet...` is a heavy object & operation. The code tries to call it when necessary.

On the other hand, I am still thinking if we should only return the first e.g. 10 names, some cluster runs 300+ VMs, expand all the names may overflow the warning window.

pkg/utils/vm.go
w13915984028: It is prepared for `func (c *CnValidator) Delete(_ *admission.Request, oldObj runtime.Object) error {` at beginning, and not used per

https://github.com/w13915984028/network-controller-harvester/blob/4db16c13a674513a3103d0d4153a0a32515fc98e/pkg/webhook/clusternetwork/validator.go#L110
```
// no need to check vmi, vm, both of them need to be stopped &/ removed related interfaces/networks when deleting nads
```

It is still kept to let vm.go is in same fashion with vmi.go

pkg/controller/manager/vlanconfig/controller.go
w13915984028: Yes, this is described in PR description:

```
2. Change the MTU of a clusternetwork
2.1 Stop all VMs attached to a specific clusternetwork
2.2 Delete/Migrate all vlanconfigs until there is a last one; otherwise change MTU from any will be denied per 1.3
2.3 Change the last vlanconfig's MTU, then it is synced to clusternetwork; if there are any NADs attached to this clusternetwork, their MTU is changed automatically
2.4 Add more vlanconfig, each needs to fill the same MTU per 2.3
```

This will forcely ensure it, to avoid user changes some of them accidentally and spend a big effort to finally figure things out.

w13915984028:
FYI: The PR is rebased with master, and will prepare for the final merge, thanks.

w13915984028:
The PR was rebased and retested per test plan https://github.com/harvester/network-controller-harvester/pull/149#issuecomment-2697974336. It is ready to merge now. Thanks for all your review.

pkg/controller/manager/vlanconfig/controller.go
```diff
@@ -83,16 +83,49 @@ func (h Handler) SetClusterNetworkUnready(_ string, vs *networkv1.VlanStatus) (*
...
+		cnCopy.Annotations[utils.KeyMTUSourceVlanConfig] = vc.Name
+		if _, err := h.cnClient.Update(cnCopy); err != nil {
+			return fmt.Errorf("failed to update cluster network %s annotation %s with MTU %s error %w", name, utils.KeyUplinkMTU, targetMTU, err)
+		}
+		klog.Infof("update cluster network %s annotation %s to %s", name, utils.KeyUplinkMTU, targetMTU)
 		return nil
 	}
 
 	// if cn is not existing
 	cn := &networkv1.ClusterNetwork{
```
ibrokethecloud, line 121: would there be a scenario where clusternetwork does not exist before a vlanconfig is created?

w13915984028: Generally, vlanconfig should not come before clusternetwork, it is cross-ensured by this PR

cn onDelete:
https://github.com/harvester/network-controller-harvester/pull/149/files#diff-702e1e8c8be9e7fd0f87527272d0fff1879b6aeb663d616640d4dcd0d17238e4R96

vlanconfig onCreate
https://github.com/harvester/network-controller-harvester/pull/149/files#diff-e30ba60a51ad3f7ae0d7a37a760cccb6c14c47fa0fa07a5b76fc1b62bd5d0112R72

If unluckily the clusternetwork is not existing (e.g. hackly removed the webhook to create a dangling vlanconfig), then L121 will create it.
https://github.com/harvester/network-controller-harvester/pull/149/files#diff-73eb21b655ff1d2a6f308f3f893c6c248f31d67f90a5d9bab4029c61e2edea31R100

PR #156, opened 2025-03-06
Remove mgmt cluster check during bridge vlan add/del

rrajendran17:
**Problem:**
During Harvester installation, the system attempts to add all VLAN IDs to the management bond/bridge. With Mellanox ConnectX-6 NICs (which support 512 hardware-offloaded VLANs and have hardware VLAN offloading enabled by default), this causes boot log warnings and prevents management VLAN 2301 from being added to the hardware offloaded VLAN list.

**Solution:**
dynamically add vlans to mgmt cluster network whenever user creates and updates the nad with mgmt cluster network
This will add vlan associated with that nad to mgmt-bo bridge in linux.

**Related Issue:**
https://github.com/harvester/harvester/issues/7650

**Test plan:**
1.Create NAD with vlan-id and attach it to cluster network mgmt
2.verify if bridge vlan show lists the vlan in the mgmt-bo
3.verify traffic between VMs connected over the vlan on mgmt-bo and its successful
4.Delete NAD with vlan-id from mgmt cluster
5.Verify if bridge vlan show removed vlan-id from mgmt-bo

This PR should be merged along with https://github.com/harvester/harvester-installer/pull/1005

### DIFF
--- a/pkg/network/iface/link.go
+++ b/pkg/network/iface/link.go
@@ -2,6 +2,7 @@ package iface
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/coreos/go-iptables/iptables"
@@ -206,4 +207,24 @@ func (l *Link) Remove() error {
 	}
 
 	return netlink.LinkDel(l)
+}
+
+func GetMgmtVlan() (vlanID int, err error) {
+	links, err := netlink.LinkList()
+	if err != nil {
+		return vlanID, err
+	}
+
+	for _, link := range links {
+		if !strings.Contains(link.Attrs().Name, "mgmt-br.") {
+			continue
+		}
+
+		result := strings.Split(link.Attrs().Name, ".")
+		if vlanID, err = strconv.Atoi(result[1]); err != nil {
+			return vlanID, err
+		}
+	}
+
+	return vlanID, nil
 }
--- a/pkg/controller/agent/nad/controller.go
+++ b/pkg/controller/agent/nad/controller.go
@@ -13,6 +13,7 @@ import (
 	"k8s.io/klog"
 
 	"github.com/harvester/harvester-network-controller/pkg/config"
+	"github.com/harvester/harvester-network-controller/pkg/network/iface"
 	"github.com/harvester/harvester-network-controller/pkg/network/vlan"
 	"github.com/harvester/harvester-network-controller/pkg/utils"
 )
@@ -22,6 +23,7 @@ const ControllerName = "harvester-network-nad-controller"
 type Handler struct {
 	nadCache  ctlcniv1.NetworkAttachmentDefinitionCache
 	nadClient ctlcniv1.NetworkAttachmentDefinitionClient
+	mgmtVlan  int
 }
 
 func Register(ctx context.Context, management *config.Management) error {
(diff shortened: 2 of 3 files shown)


rrajendran17:
Note: since custom vlan added to mgmt-bo during harvester installation should not be removed at any point of time, I am planning to add code changes during Register() of nad controller and fetch and store the custom vlan configured by user during installation.
And a check could be added during RemoveLocalArea() for that vid and it will be removed from mgmt-bo.

I am yet to add those changes to this PR.Please let me know if that looks fine or any suggestions for better way to do this ?

w13915984028 (dismissed):
LGTM, thanks.

Meanwhile, do you have a place to record which VID is from mgmt installation? When adding other vm network/storagenetwork on mgmt network, a check is needed to avoid conflict with this VID.

At the moment, the mgmt clusternetwork has no vlanconfig, and the installation configuration is not updated to clusternetwork/vlanconfig. New enhancements are needed to unify the mgmt clusternetwork same as others.  E.g. a node controller converts mgmt network related config into a vlanconfig.

There are some comments and related specific processing on mgmt network on PR https://github.com/harvester/network-controller-harvester/pull/149.

rrajendran17:
> LGTM, thanks.
> 
> Meanwhile, do you have a place to record which VID is from mgmt installation? When adding other vm network/storagenetwork on mgmt network, a check is needed to avoid conflict with this VID.
> 
> At the moment, the mgmt clusternetwork has no vlanconfig, and the installation configuration is not updated to clusternetwork/vlanconfig. New enhancements are needed to unify the mgmt clusternetwork same as others. E.g. a node controller converts mgmt network related config into a vlanconfig.
> 
> There are some comments and related specific processing on mgmt network on PR #149.

Yes I am planning this as mentioned in https://github.com/harvester/network-controller-harvester/pull/156#issuecomment-2702467664. Please let me know if there are some better options within network-controller for now until we go with above enhancement.

rrajendran17:
I have added changes for storing the mgmt vlan configured during installation and added checks so that we do not remove this vlan from mgmt-bo bridge accidentally.  

@w13915984028 @starbops @mingshuoqiu Please review whenever you get time, Thanks

pkg/network/iface/link.go
```diff
@@ -207,3 +208,23 @@ func (l *Link) Remove() error {
...
 }
+
+func GetMgmtVlan() (vlanID int, err error) {
+	links, err := netlink.LinkList()
+	if err != nil {
+		return vlanID, err
+	}
+
+	for _, link := range links {
+		if !strings.Contains(link.Attrs().Name, "mgmt-br.") {
```
w13915984028, line 219: there are two consts to construct `mgmt-br`:
ManagementClusterNetworkName
BridgeSuffix

pkg/network/iface/link.go
```diff
@@ -207,3 +208,23 @@ func (l *Link) Remove() error {
...
+			continue
+		}
+
+		result := strings.Split(link.Attrs().Name, ".")
+		if vlanID, err = strconv.Atoi(result[1]); err != nil {
+			return vlanID, err
+		}
+	}
+
+	return vlanID, nil
```
w13915984028, line 229: if mgmt does not configure a VID, will this function:

return 0, nil ?

w13915984028 (commented):
need to check the upgrade path:

when PODs are upgraded, but NODEs are still not,  will the new controller work on legacy node ?

pkg/network/iface/link.go
rrajendran17: yes, do you think we need to configure mgmt vlan with vid 1 in that case ?

pkg/network/iface/link.go
rrajendran17: sure, will change them

PR #163, opened 2025-03-28
Support virtual network

rrajendran17:
**Problem:**
Support virtual network using kubeovn

**Solution:**
1.Since kube-ovn does not support linux bridge, we need to skip the following in network-controller
       a.skip webhook validations related to bridge
       b.skip controller logic involving bridge and route connectivity checks.
       c.Route connectivity checks are needed for checking connectivity between harvester node and vlan network,
         Since no vlan is involved with kube-ovn NAD and mgmt cluster is used for underlay, route connectivity check could be 
         skipped for now.

2.Integrate kubeovn apis for subnet and vpcs into network controller

3.Add webhooks for subnets,vpcs for validations.

**Related Issue:**
https://github.com/harvester/harvester/issues/7332

**Test plan:**

https://github.com/harvester/charts/pull/354 for rbac


mingshuoqiu (dismissed):
The commit mainly integrates kube-ovn api,  introduces subnet and vpc validator and check the NAD provider type against `kube-ovn`.
LGTM. Thanks

pkg/webhook/subnet/validator.go
```diff
@@ -0,0 +1,151 @@
...
+
+	subnetSpec := subnet.Spec
+
+	if subnetSpec.Provider == "" {
+		return fmt.Errorf("provider is empty for subnet %s in %s", subnet.Name, subnet.Namespace)
+	}
+
+	if subnetSpec.CIDRBlock == "" || subnetSpec.Gateway == "" {
+		return fmt.Errorf("subnet cidr/gw is empty for subnet %s in %s", subnet.Name, subnet.Namespace)
+	}
```
starbops, line 45: Do we want to allow empty-gateway subnets? The if statement and the error message seem not aligned.

pkg/webhook/subnet/validator.go
```diff
@@ -0,0 +1,151 @@
...
+
+	err = v.checkifVpcExists(newSubnetSpec.Vpc)
+	if err != nil {
+		return fmt.Errorf("vpc does not exists for subnet %s in %s err=%v", newSubnet.Name, newSubnet.Namespace, err)
+	}
+
+	return nil
+}
+
+func (v *Validator) checkifNadExists(subnetProvider string) error {
```
starbops, line 104:
```suggestion
func (v *Validator) checkIfNadExists(subnetProvider string) error {
```

pkg/webhook/subnet/validator.go
```diff
@@ -0,0 +1,151 @@
...
+	providerName := strings.Split(subnetProvider, ".")
+	_, err := v.nadCache.Get(providerName[1], providerName[0])
+	if err != nil {
+		return fmt.Errorf("failed to get nad %s %s created for the subnet", providerName[1], providerName[0])
+	}
+
+	return nil
+}
+
+func (v *Validator) checkifVpcExists(subnetVpc string) error {
```
starbops, line 114:
```suggestion
func (v *Validator) checkIfVpcExists(subnetVpc string) error {
```

pkg/webhook/subnet/validator.go
```diff
@@ -0,0 +1,151 @@
...
+	if err != nil {
+		return fmt.Errorf("vpc does not exists for subnet %s in %s err=%v", newSubnet.Name, newSubnet.Namespace, err)
+	}
+
+	return nil
+}
+
+func (v *Validator) checkifNadExists(subnetProvider string) error {
+	providerName := strings.Split(subnetProvider, ".")
+	_, err := v.nadCache.Get(providerName[1], providerName[0])
```
starbops, line 106: Might need some checks here to ensure we have at least two elements.

pkg/webhook/subnet/validator.go
```diff
@@ -0,0 +1,151 @@
...
+		return fmt.Errorf("failed to get nad %s %s created for the subnet", providerName[1], providerName[0])
+	}
+
+	return nil
+}
+
+func (v *Validator) checkifVpcExists(subnetVpc string) error {
+	if subnetVpc != "" {
+		_, err := v.vpcCache.Get(subnetVpc)
+		if err != nil {
```
starbops, line 117: It would be better if we can separate different error types so that we can provide clear information to the user. For instance, https://github.com/harvester/network-controller-harvester/blob/f9304a1f64914ad989452927a79fe35fba0509bd/pkg/webhook/vlanconfig/validator.go#L186

pkg/webhook/subnet/validator.go
```diff
@@ -0,0 +1,151 @@
...
+		return fmt.Errorf("vpc does not exists for subnet %s in %s err=%v", newSubnet.Name, newSubnet.Namespace, err)
+	}
+
+	return nil
+}
+
+func (v *Validator) checkifNadExists(subnetProvider string) error {
+	providerName := strings.Split(subnetProvider, ".")
+	_, err := v.nadCache.Get(providerName[1], providerName[0])
+	if err != nil {
```
starbops, line 107: ditto

starbops (commented):
Thanks for the PR. I have a general question regarding the validators presented here. It looks like our validator and kube-ovn both implement duplicate checks. We would probably only need to add those that are needed to implement our own business logic.

PR #166, opened 2025-04-08
Remove mgmt cluster check during bridge vlan add/del

rrajendran17:
**Problem:**
During Harvester installation, the system attempts to add all VLAN IDs to the management bond/bridge. With Mellanox ConnectX-6 NICs (which support 512 hardware-offloaded VLANs and have hardware VLAN offloading enabled by default), this causes boot log warnings and prevents management VLAN 2301 from being added to the hardware offloaded VLAN list.

**Solution:**
Dynamically add vlans to mgmt cluster network whenever user creates and updates the nad with mgmt cluster network
This will add vlan associated with that nad to mgmt-bo bridge in linux.

**Related Issue:**
https://github.com/harvester/harvester/issues/7650

**Test plan:**
1.Create NAD with vlan-id and attach it to cluster network mgmt
2.verify if bridge vlan show lists the vlan in the mgmt-bo
3.verify traffic between VMs connected over the vlan on mgmt-bo and its successful
4.Delete NAD with vlan-id from mgmt cluster
5.Verify if bridge vlan show removed vlan-id from mgmt-bo

This PR should be merged along with https://github.com/harvester/harvester-installer/pull/1005

Upgrade scenario: (new network controller pods running on old node)
1.New network controller pod retrieves the mgmt vlan and stores it during init.
   (no change in fetching the mgm-br or mgmt-br.vlanid interface from old node)
2.New network controller pod will try to add vlan id to the mgmt-bo if there is a user configured NAD with vlan id under mgmt cluster.Since the old node already has all vlans added, this vid addition will be no-op in kernel.
Note:Assumption there be no user other user operations on the NAD for delete during upgrade

### DIFF
--- a/pkg/network/iface/link.go
+++ b/pkg/network/iface/link.go
@@ -2,6 +2,7 @@ package iface
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/coreos/go-iptables/iptables"
@@ -206,4 +207,34 @@ func (l *Link) Remove() error {
 	}
 
 	return netlink.LinkDel(l)
+}
+
+func GetMgmtVlan() (vlanID int, err error) {
+	links, err := netlink.LinkList()
+	if err != nil {
+		return vlanID, err
+	}
+
+	mgmtBrIntf := utils.ManagementClusterNetworkName + "-" + BridgeSuffix + "."
+	for _, link := range links {
+		if !strings.HasPrefix(link.Attrs().Name, mgmtBrIntf) {
+			continue
+		}
+
+		result := strings.Split(link.Attrs().Name, ".")
+		if len(result) < 2 {
+			return vlanID, fmt.Errorf("invalid link name format: %s", link.Attrs().Name)
+		}
+
+		if vlanID, err = strconv.Atoi(result[1]); err != nil {
+			return vlanID, err
+		}
+	}
+
+	if vlanID != 0 {
+		return vlanID, nil
+	}
+
+	//return default vid=1
+	return 1, nil
 }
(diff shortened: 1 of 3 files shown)


rrajendran17:
@w13915984028 My understanding is upgrade scenario with this changes should not cause any problem.

Upgrade scenario: (new network controller pods running on old node)
1.New network controller pod retrieves the mgmt vlan and stores it during init.
(no change in fetching the mgm-br or mgmt-br.vlanid interface from old node)
2.New network controller pod will try to add vlan id to the mgmt-bo if there is a user configured NAD with vlan id under mgmt cluster.Since the old node already has all vlans added, this vid addition will be no-op in kernel.

Note:Assumption there be no user other user operations on the NAD for delete during upgrade

pkg/network/iface/link.go
```diff
@@ -207,3 +208,33 @@ func (l *Link) Remove() error {
...
+
+func GetMgmtVlan() (vlanID int, err error) {
+	links, err := netlink.LinkList()
+	if err != nil {
+		return vlanID, err
+	}
+
+	mgmtBrIntf := utils.ManagementClusterNetworkName + "-" + BridgeSuffix + "."
+	for _, link := range links {
+		if !strings.Contains(link.Attrs().Name, mgmtBrIntf) {
```
w13915984028, line 220: `HasPrefix` is even safe

w13915984028 (approved):
LGTM, thanks.

w13915984028 (approved):
LGTM, thanks.

mingshuoqiu (approved):
LGTM. Thanks

PR #173, opened 2025-06-11
nad changes for type kube-ovn and subnet wehook validations

rrajendran17:
**Problem:**
Support for new type kube-ovn in network controller

**Solution:**
Creation of Network Attachment Definition with type kube-ovn (OverlayNetwork)
- nad webhook must be modified to accept type "kube-ovn" and without any bridge name.
-  nad controller updates the nad cache label with new network type "OverlayNetwork"
- nad controller logic must skip processing related to vlan and checking L3 network connectivity from DHCP.

Subnet webhook validations:
- check if NAD exists during subnet create/update
- check if vpc exists during subnet create/update

**Related Issue:**
https://github.com/harvester/harvester/issues/7332

**Test plan:**


pkg/webhook/nad/mutator.go
```diff
@@ -105,6 +105,17 @@ func (m *Mutator) ensureLabels(nad *cniv1.NetworkAttachmentDefinition, oldConf,
...
+	if newConf.Type == utils.CNITypeKubeOVN {
+		labels[utils.KeyNetworkType] = string(utils.OverlayNetwork)
+		labels[utils.KeyNetworkReady] = utils.ValueTrue
+		return admission.Patch{
+			admission.PatchOp{
+				Op:    admission.PatchOpReplace,
+				Path:  "/metadata/labels",
+				Value: labels,
+			}}, nil
+	}
```
starbops, line 117: If we want to short-circuit the `ensureLabels` function for kube-ovn type NADs, could we move the snippet before the VLAN-check one?

pkg/webhook/nad/validator.go
```diff
@@ -131,6 +131,24 @@ func (v *Validator) checkNadConfig(bridgeConf *utils.NetConf, nad *cniv1.Network
 		return fmt.Errorf("config is empty")
 	}
 
+	//skip bridge config validation for type kube-ovn
+	if bridgeConf.Type == utils.CNITypeKubeOVN {
```
starbops, line 135: nit: maybe it's time to rename the variable to something more general, such as `nadConf`, instead of `bridgeConf`.

pkg/webhook/nad/validator.go
```diff
@@ -131,6 +131,24 @@ func (v *Validator) checkNadConfig(bridgeConf *utils.NetConf, nad *cniv1.Network
...
+		if bridgeConf.Provider == "" {
+			return fmt.Errorf("provider is empty for cni type %s", utils.CNITypeKubeOVN)
+		}
+
+		providerName := strings.Split(bridgeConf.Provider, ".")
+		if len(providerName) < 2 {
+			return fmt.Errorf("invalid provider length %d for provider %s", len(providerName), bridgeConf.Provider)
+		}
```
starbops, line 143: What's the rationale here? Should it be `3` instead of `2` (`<name>.<namespace>.ovn`)?

pkg/webhook/nad/validator.go
```diff
@@ -131,6 +131,24 @@ func (v *Validator) checkNadConfig(bridgeConf *utils.NetConf, nad *cniv1.Network
...
+
+		if bridgeConf.Name != providerName[0] {
+			return fmt.Errorf("invalid provider name %s for nad %s", providerName[0], bridgeConf.Name)
+		}
```
starbops, line 147: Is it mandatory for the config name to equal to the provider name? IIRC, the provider name should be the same as the NAD name.

BTW, do we want to check the namespace as well?

pkg/webhook/subnet/validator.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	return nil
+}
+
+func (v *Validator) checkIfNadExists(subnetProvider string) (err error) {
+	providerName := strings.Split(subnetProvider, ".")
+	if len(providerName) < 2 {
+		return fmt.Errorf("invalid provider for subnet %s", subnetProvider)
+	}
```
starbops, line 93: Ditto

pkg/webhook/nad/validator.go
```diff
@@ -131,6 +131,24 @@ func (v *Validator) checkNadConfig(bridgeConf *utils.NetConf, nad *cniv1.Network
...
+		if bridgeConf.Provider == "" {
+			return fmt.Errorf("provider is empty for cni type %s", utils.CNITypeKubeOVN)
+		}
+
+		providerName := strings.Split(bridgeConf.Provider, ".")
+		if len(providerName) < 2 {
+			return fmt.Errorf("invalid provider length %d for provider %s", len(providerName), bridgeConf.Provider)
+		}
+
+		if bridgeConf.Name != providerName[0] {
```
ibrokethecloud, line 145: name is an optional field in the spec

pkg/webhook/nad/mutator.go
rrajendran17: We will need the KeyLastNetworkType label if we are changing between different network types (kubeovn to l2vlan/untagged and vice-versa)

pkg/webhook/nad/validator.go
rrajendran17: yeah right, it is a mistake

pkg/webhook/nad/validator.go
rrajendran17: I see nad spec config derives the name from nad, but better to use nad's name and namespace

pkg/webhook/nad/validator.go
```diff
@@ -200,6 +218,11 @@ func (v *Validator) checkNadConfigBridgeName(oldNC, newNC *utils.NetConf) error
 		return fmt.Errorf("new nad config is empty")
 	}
```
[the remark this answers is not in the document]
rrajendran17: I think it is ok to change between network types from kube-ovn to l2vlannetwork/untagged and vice-versa.
Currently we allow updates on nad for the network types when VMs are stopped in state.
Later they come up and use the new network type.

I added and tested those part of test steps.

pkg/controller/manager/nad/controller.go
```diff
@@ -208,14 +210,19 @@ func (h Handler) ensureLabels(nad *cniv1.NetworkAttachmentDefinition) error {
 		return err
 	}
 
-	if netconf.Vlan != 0 {
-		labels[utils.KeyNetworkType] = string(utils.L2VlanNetwork)
-		labels[utils.KeyVlanLabel] = strconv.Itoa(netconf.Vlan)
+	if netconf.Type == utils.CNITypeKubeOVN {
+		labels[utils.KeyNetworkType] = string(utils.OverlayNetwork)
+		cnName = utils.ManagementClusterNetworkName
```
w13915984028, line 215: 1. Does it mean: the kubeovn nad only supports on top of mgmt network ATM ?

```
cnName =labels[utils.KeyClusterNetworkLabel]
if cnName == ""
  cnName = utils.ManagementClusterNetworkName  // fallback or report error
```

The normal creation process is: user fills the NAD fields from UI, UI/harvester mutator should have added the required info like  network type, cluster network name

2. Need to double check this block

on `bridge` mode NAD, the missing labels  are not a big problem, the nad itself can deduce clusternetwork, the `br-*` is the key word.

on `overlay` mode,  those labels must be ensured from the UI/kubectl input yaml,  the webhook needs to validate: if nad is overlay, the related clusternetwork  label must be placed and the clusternetwork is trully existing

pkg/webhook/nad/mutator.go
```diff
@@ -100,11 +100,28 @@ func (m *Mutator) ensureLabels(nad *cniv1.NetworkAttachmentDefinition, oldConf,
...
 	}
 
+	if oldConf.Vlan != 0 && oldConf.Vlan != newConf.Vlan {
+		labels[utils.KeyLastVlanLabel] = strconv.Itoa(oldConf.Vlan)
+	}
+
+	if newConf.Type == utils.CNITypeKubeOVN {
+		labels[utils.KeyNetworkType] = string(utils.OverlayNetwork)
+		labels[utils.KeyNetworkReady] = utils.ValueTrue
+		labels[utils.KeyClusterNetworkLabel] = utils.ManagementClusterNetworkName
```
w13915984028, line 115: labels[utils.KeyClusterNetworkLabel] = utils.ManagementClusterNetworkName
->
if labels[utils.KeyClusterNetworkLabel] == "" {
  labels[utils.KeyClusterNetworkLabel] = utils.ManagementClusterNetworkName
}

w13915984028 (commented):
A few questions, thanks.

It seems the ovn is only on top of mgmt CN ATM, and this caused a top-level info is missing:

The `clusternetwork` needs to have a attribute/filed to mark it as the `overlay` / `bridge`(default) network; and add related `vlanconfig` (not accurate) to configurate the `uplink`' physical link &/ IP(most important); thus to establish the overlay network first.  Using mgmt skips this step, but a design and related info is still required.  In real deaployment, the so called `data-plane` clusternetwork is essential to isolate the VM's network from the mgmt network.

When creating nad, it needs the parent `clusternetwork` info, and the nad overlay attribute is cross checked with `clusternetwork`'s overlay attribute.

When updating nad, it should not switch between the `overlay` mode and the exsting `bridge` mode.  This has been discussed with @rrajendran17 offline.

even the bridge mode nad, switch clusternetwork (namely change the bridge name) is also not allowed.

pkg/controller/manager/nad/controller.go
rrajendran17: Yes for now, overlay type only supports mgmt network.
We thought from UI either we make sure to use only mgmt or it can be empty.
nad webhook also validates this and nad mutator fills the cluster network as mgmt for type kube-ovn.

rrajendran17: True, in future when overlay type is supported on cluster network other than mgmt, this code needs to be refactored and somehow cluster network name must be available for nad mutator from yaml to perform the similar logic we do for L2VlanNetwork.

pkg/webhook/nad/validator.go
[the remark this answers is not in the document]
rrajendran17: Restricted the network type change bridge to ovn and vice-versa

rrajendran17:
Currently restricting the network type change from bridge to ovn and vice-versa and the PR takes care of those changes.

pkg/controller/manager/nad/controller.go
w13915984028: BTW, Harvester supports [witness node](https://docs.harvesterhci.io/v1.5/advanced/witness), which does not run any workload, is it excluded from the overlay network setup and skipped to deploy those daemons?

w13915984028 (dismissed):
LGTM, thanks.

mingshuoqiu (dismissed):
LGTM. Thanks

ibrokethecloud (dismissed):
lgtm. thanks.

w13915984028 (approved):
LGTM, thanks.

starbops:
@mergify backport v1.6

PR #174, opened 2025-06-16
Log version info when app starts

w13915984028:
**Problem:**

commit ID is not logged

**Solution:**

**Related Issue:**
https://github.com/harvester/harvester/issues/8486

**Test plan:**

network related POD log has info like

```
Starting harvester-network-controller version acb9fc55
```


cmd/network-controller/main.go
```diff
@@ -98,17 +104,11 @@ func run(c *cli.Context, registerFuncList []config.RegisterFunc, leaderelection
 	helperImage := c.String("helper-image")
 
 	if threadiness <= 0 {
-		klog.Infof("Can not start with thread count of %d, please pass a proper thread count.", threadiness)
-		return nil
+		klog.Infof("Thread count of %d is invalid, fallback to default value 2.", threadiness)
+		threadiness = 2
```
ihcsim, line 108: Nit: consider making it a const since it's also used in the CLI flag definition.

w13915984028: Good idea, updated.

PR #182, opened 2025-07-01
[WIP]Enh766 supports vlan trunk mode nad

w13915984028:
**Problem:**

**Solution:**

1. Summarize the vlan ID set into a separate CRD object
2. Segregate the functions into two parts: 
- nad controller: update the vid set
- vlan config controller: update the vid set to Linux driver (`bridge vlan show` result) 

**Related Issue:**

**Test plan:**


w13915984028:
replaced by https://github.com/harvester/network-controller-harvester/pull/186 ATM.

PR #186, opened 2025-07-07
Support vlan trunk nad

w13915984028:
**Problem:**

Harvester network controller does not support vlan trunk mode nad

**Solution:**

    Support vlan trunk mode nad
    
     Add new vidset to represent the 4K vlan set, can append, diff
     Massive refactor to optimize and support vlan trunk
     Remove the silently shortcuted route add/remove
     Break the loop importing with Harvester
     Fix a few bugs
     Remove the nad-agent controller

A document for referrence: https://github.com/w13915984028/harvester-develop-summary/blob/main/test-harvester-vlan-trunk-mode-nad.md

**main changes:**

**network-controller-manager:**
  1. when any nad is changed, re-compute the current cluster network's vlan set, if any is added/removed, update the cluster network object

**network-controller-agent:**
  1. when cluster network changes, fetching the existing vlans, compare it with the vlan set computed via nad list, get the added list and removed list;  if `up-link` is not existing, skip
  2. remove nad controller on agent
  3. vlan config controller:  on change: is only responsible for setup bridge & up-link, then trigger cn change;  on remove: remove up-link && bridge

  4. remove the `route` related code, it has no effect

**reason to drop nad-agent-controller**

nad is not a node-level object, but nad-agent-controller is, when nad-agent-controller updates nad object, race can happen.

```
                                         nad agent
                              node 1         node 2          node 3
      nad1                      |              |               |
    vid 100->200                |              |               |        nad1 vid changes from 100 to 200
        ----------------------->--------------->--------------->        change is synced to all node cache
                           remove old 100     rebooted         |        node 1 handle nad1, node 2 agent is rebooted
          update, remove        |              |               |        node 1 update nad1 to remove KeyLastVlanLabel
        <-----------------------|              |               |  
           KeyLastVlanLabel     |              |               |
                                |              |               |        
        ----------------------->--------------->--------------->        when nad agent on node 2 restarts,
                                                                        it does not see nad1 has the KeyLastVlanLabel
 
based on the processing sequences, nad-agent on node 3 might also only see the nad1 without label KeyLastVlanLabel
```

**enhance later**
1. Fully drop the `cidr` part when collecting vlan info
2. Support single vlan mode VlanIDSet to improve performance

**Related Issue:**

**Test plan:**

Per: https://github.com/harvester/network-controller-harvester/pull/186#issuecomment-3079976438

**additional context**

The last failure case on issue https://github.com/harvester/harvester/issues/7676#issuecomment-3077879194, is also addressed in this PR

The  failure case on https://github.com/harvester/harvester/issues/7676 , is a long-existing issue of current network-controller-harvester, even in earlier version

specifically, on those lines
https://github.com/harvester/network-controller-harvester/blob/4ec00f753ff2d38ca4664e4dd8d53418c7ee7266/pkg/controller/agent/nad/controller.go#L57

https://github.com/harvester/network-controller-harvester/blob/4ec00f753ff2d38ca4664e4dd8d53418c7ee7266/pkg/controller/agent/nad/controller.go#L186

the agent-controller will remove the nad's last vid when it exists on the nad, and then amend the nad,

```
func (h Handler) OnChange(_ string, nad *nadv1.NetworkAttachmentDefinition) (*nadv1.NetworkAttachmentDefinition, error) {
	if nad == nil || nad.DeletionTimestamp != nil {
		return nil, nil
	}

	klog.Infof("nad configuration %s/%s has been changed: %s", nad.Namespace, nad.Name, nad.Spec.Config)

	// nadCopy could be nil
	nadCopy, err := h.removeOutdatedLocalArea(nad)
	if err != nil {
		return nil, fmt.Errorf("remove outdated local area for nad %s/%s failed, error: %w", nad.Namespace, nad.Name, err)
	}
```

however, **the assumption has some errors**:
(1) not every node is added in certain cluster-network, namely, agent should not process nad blindly
(2) one agent updates the nad, might cause another node agent has no chance to amend to last vid.

It is fixed in this PR.


w13915984028:
About the current `route` related code

(1) The designed process to add vlan&cidr route

nad add  https://github.com/harvester/network-controller-harvester/blob/4ec00f753ff2d38ca4664e4dd8d53418c7ee7266/pkg/controller/agent/nad/controller.go#L63
  add vlan & cidr https://github.com/harvester/network-controller-harvester/blob/4ec00f753ff2d38ca4664e4dd8d53418c7ee7266/pkg/controller/agent/nad/controller.go#L118
                  https://github.com/harvester/network-controller-harvester/blob/4ec00f753ff2d38ca4664e4dd8d53418c7ee7266/pkg/network/vlan/vlan.go#L110
  ensure route via gateway  https://github.com/harvester/network-controller-harvester/blob/4ec00f753ff2d38ca4664e4dd8d53418c7ee7266/pkg/network/iface/route.go#L73
  

(2) Add a nad and the debug log:
```
I0714 14:08:30.530631       1 controller.go:54] nad configuration default/nad3 has been changed: {"cniVersion":"0.3.1","name":"nad3","type":"bridge","bridge":"cn2-br","promiscMode":true,"vlan":0,"vlanTrunk":[{"minID":3100,"maxID":3104}],"ipam":{},"mtu":1600}
I0714 14:08:30.530644       1 controller.go:54] nad configuration default/nad30 has been changed: {"cniVersion":"0.3.1","name":"nad30","type":"bridge","bridge":"cn2-br","promiscMode":true,"vlan":30,"ipam":{},"mtu":1600}
I0714 14:08:30.530630       1 controller.go:104] lm mgmt matches node harv41
I0714 14:08:30.533772       1 route.go:39] get all routes 56
I0714 14:08:30.533800       1 route.go:41]  R: {Ifindex: 4 Dst: 0.0.0.0/0 Src: <nil> Gw: 192.168.122.1 Flags: [] Table: 254 Realm: 0}
I0714 14:08:30.533809       1 route.go:41]  R: {Ifindex: 6 Dst: 10.52.0.2/32 Src: <nil> Gw: <nil> Flags: [] Table: 254 Realm: 0}
...
I0714 14:08:30.534129       1 route.go:41]  R: {Ifindex: 67 Dst: 10.52.0.58/32 Src: <nil> Gw: <nil> Flags: [] Table: 254 Realm: 0}
I0714 14:08:30.534135       1 route.go:41]  R: {Ifindex: 4 Dst: 192.168.122.0/24 Src: 192.168.122.141 Gw: <nil> Flags: [] Table: 254 Realm: 0}
I0714 14:08:30.534141       1 route.go:72] route for cidr 192.168.30.0/24, ip 192.168.30.0 network 192.168.30.0/24 gateway <nil> 
I0714 14:08:30.534199       1 route.go:80] route: {Ifindex: 4 Dst: 192.168.30.0/32 Src: 192.168.122.141 Gw: 192.168.122.1 Flags: [] Table: 254 Realm: 0}
I0714 14:08:30.534216       1 route.go:82] it has the route {Ifindex: 4 Dst: 192.168.30.0/32 Src: 192.168.122.141 Gw: 192.168.122.1 Flags: [] Table: 254 Realm: 0}


I0714 14:08:30.534228       1 controller.go:54] nad configuration default/nad40 has been changed: {"cniVersion":"0.3.1","name":"nad40","type":"bridge","bridge":"cn2-br","promiscMode":true,"vlan":40,"ipam":{},"mtu":1600}



harv41:/home/rancher # ip route
default via 192.168.122.1 dev mgmt-br proto dhcp 
10.52.0.2 dev calicbb5b2c1f3f scope link 
...
10.52.0.58 dev cali833c0c52299 scope link 
192.168.122.0/24 dev mgmt-br proto kernel scope link src 192.168.122.141 
```

(3) The default route

On Harvester node, there is always a default route https://github.com/harvester/harvester-installer/pull/876; the current source code looks always return from below path, no real route like `dest 192.168.30.0/32 gw 192.168.30.1` is added.

```
// EnsureRouteViaGateway will add the route via gateway if not existing
func EnsureRouteViaGateway(cidr string) error {
	if cidr == "" {
		return nil
	}

	ip, network, err := net.ParseCIDR(cidr)
	if err != nil {
		return err
	}

	gateway, linkIndex, err := getGateway()
	if err != nil {
		return err
	}

	klog.Infof("route for cidr %s, ip %+v network %+v current gateway %+v ", cidr, ip, network, gateway)

	routes, err := netlink.RouteGet(ip)
	if err != nil {
		return err
	}

	for _, route := range routes {
		klog.Infof("route: %s", route.String())
		if route.Gw != nil {
			klog.Infof("it has the route %s", route.String())
			return nil  // return
		}
	}
```

(4) legaycy code was aeed via below commit and PR:
                                
https://github.com/harvester/network-controller-harvester/commit/84bce4c33f17670d991abbbc2287dbb5f6a687d9#diff-e531d083e7bcfb716feb4305695b3db455be23bf77b1593855fdb7c4859ae25b

Add routes from hosts to vlan network
- Add the annotation `network.harvesterhci.io/layer3-network` to nad CR. There are two ways to get the layer 3 network information.
  - `auto`: Create a job to set up a network helper pod to get th CIDR and gateway through the DHCP discovery.
  - `manual`: Users should configure the CIDR and gateway manually. It's the default option.
- Ping the gateway of the VLAN network recorded in the annotation periodically to check the connectivity from host to the VLAN network. The controller will record the connectivity in the annotation as well.
- Add the route to reach the addresses in the CIDR via the default gateway on every host.

note:

https://github.com/harvester/network-controller-harvester/commit/84bce4c33f17670d991abbbc2287dbb5f6a687d9#diff-33b9dc9742296ae67bc18fa3b0f5c743e0f38fd7721c8720e83234af7e56f7f0R80

v, err := vlan.GetVlan(h.mgmtNetwork)

https://docs.harvesterhci.io/v1.0/networking/harvester-network#enabling-default-vlan-network

legacy, the `vlan network` can share the `uplink` with mgmt, and there is only one additional vlan network.

current: `mgmt` and `vlan` network are separated with uplink


(5) additional context

https://github.com/harvester/network-controller-harvester/commit/84bce4c33f17670d991abbbc2287dbb5f6a687d9#diff-e531d083e7bcfb716feb4305695b3db455be23bf77b1593855fdb7c4859ae25b

Add routes from hosts to vlan network
- Add the annotation `network.harvesterhci.io/layer3-network` to nad CR. There are two ways to get the layer 3 network information.
  - `auto`: Create a job to set up a network helper pod to get th CIDR and gateway through the DHCP discovery.
  - `manual`: Users should configure the CIDR and gateway manually. It's the default option.
- Ping the gateway of the VLAN network recorded in the annotation periodically to check the connectivity from host to the VLAN network. The controller will record the connectivity in the annotation as well.
- Add the route to reach the addresses in the CIDR via the default gateway on every host.

https://github.com/harvester/network-controller-harvester/pull/48 https://github.com/harvester/network-controller-harvester/pull/48     Nov 24, 2021
https://github.com/harvester/harvester/issues/1476 [FEATURE] Add network reachability detection from host for the VLAN network #1476
https://github.com/harvester/harvester/issues/1474 [Feature][Bug] Add a route to prove RKE1/RKE2 in VLAN network can access the VIP

https://github.com/harvester/harvester/issues/1059 [Enhancement] SSH connectivity via the public IP fails when VM has both mgmt network and public network #1059
https://github.com/harvester/harvester/issues/1424 [BUG/ENHANCEMENT] VIP configured in a VLAN network can not be reached

https://github.com/harvester/network-controller-harvester/commit/de19d74cd71609ed221446d14aed6a301c49eed7
Ensure iptables forward rules for the VLAN network

https://github.com/harvester/network-controller-harvester/commit/e5c1d9e8bc4639bb087f41b9c655c4d8995417dd https://github.com/harvester/network-controller-harvester/pull/28
keep default route after address changed

https://github.com/harvester/network-controller-harvester/commit/e73a7dd399cd0ef622cd2a1fad7d04a1a2ac928a#diff-b533c060ee06452f6e0291e2c3443e639c7fcc18ef66e0b6020d533d792929f7 https://github.com/harvester/network-controller-harvester/pull/25

https://github.com/harvester/harvester/issues/6671 [ENHANCEMENT] Add additional info for Route Connectivity "Inactive" Status
https://github.com/harvester/harvester/issues/3037 [ENHANCEMENT] Guest cluster VLAN should not require connectivity to the Harvester VIP #3037

pkg/utils/crd/crd.go
```diff
@@ -0,0 +1,226 @@
...
+
+// cloned from pkg/harvester/util/crd/init.go, to break loop importing
+
+// FromGV returns an abstract namespace-scope CRD handler via group version and kind.
+func FromGV(gv schema.GroupVersion, kind string, obj interface{}) crd.CRD {
+	return crd.FromGV(gv, kind).WithSchemaFromStruct(obj)
+}
+
+// NonNamespacedFromGV returns a cluster-scope CRD abstract handler via group version and kind.
+func NonNamespacedFromGV(gv schema.GroupVersion, kind string, obj interface{}) crd.CRD {
```
ibrokethecloud, line 32: this seems to be unused

pkg/controller/manager/clusternetwork/controller.go
```diff
@@ -144,10 +144,16 @@ func (h Handler) setNadReadyLabel(cn *networkv1.ClusterNetwork) error {
...
+		if nad.DeletionTimestamp != nil {
+			continue
+		}
 		if nad.Labels[utils.KeyNetworkReady] == isReady {
 			continue
 		}
 		nadCopy := nad.DeepCopy()
+		if nadCopy.Labels == nil {
+			nadCopy.Labels = make(map[string]string)
+		}
```
ibrokethecloud, line 156: should we move this after the check for `nad.DeletionTimestamp` since we are checking 
```
if nad.Labels[utils.KeyNetworkReady] == isReady {
			continue
		}
```

if nad.Labels are not going to be empty, then nadCopy is going to inherit the labels too.

```suggestion
		if nad.Labels == nil {
			nad.Labels = make(map[string]string)
		}
```

pkg/network/vlan/vlan.go
```diff
@@ -98,7 +101,7 @@ func (v *Vlan) Teardown() error {
 	return nil
 }
 
-func (v *Vlan) AddLocalArea(la *LocalArea) error {
+func (v *Vlan) AddLocalArea(la *utils.LocalArea) error {
```
ibrokethecloud, line 104: seems like an unused method, as vlan controller is calling `AddLocalAreas`

pkg/utils/vlan.go
```diff
@@ -0,0 +1,217 @@
...
+	if vid < MinVlanID || vid > MaxVlanID {
+		return fmt.Errorf("vlan %v is out of range [%v .. %v]", vid, MinVlanID, MaxVlanID)
+	}
+	// 0 is always skipped
+	if vid == 0 {
+		return nil
+	}
+
+	if len(vis.vidSet) < (vid + 1) {
+		fmt.Errorf("vlan set length is %v, not enough space to store vid %v", len(vis.vidSet), vid)
```
ibrokethecloud, line 67: likely missing a return

pkg/controller/agent/vlanconfig/controller.go
```diff
@@ -134,7 +136,7 @@ func (h Handler) initialize() error {
 
 // MatchNode will also return the executed vlanconfig with the same clusterNetwork on this node if existing
 func (h Handler) MatchNode(vc *networkv1.VlanConfig) (bool, error) {
-	if vc.Annotations == nil || vc.Annotations[utils.KeyMatchedNodes] == "" {
+	if vc.Annotations[utils.KeyMatchedNodes] == "" {
```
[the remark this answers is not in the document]
w13915984028: This is on the fact that golang map can `read`  an element even when the map itself is nil, to make the code to be a bit compacted.

```
	var ss map[string]string
	if ss["a"] == "" {
		fmt.Println("no element")
	}
```

pkg/utils/crd/crd.go
w13915984028: the crd/crd.go is copied from Harvester and kept unchanged & same with Harvester,  mainly for breaking the loop-importing between Harvester and network-controller

this function has no caller at the moment, but will be kept to be on sync with Harvester

rrajendran17:
@w13915984028 looks like there is a problem when mgmt interface is on a custom vlan.
The custom vlan gets removed on the mgmt-bo due to which connectivity is lost to the VIP.
IIRC, We need to retain the custom vid configured on the mgmt both on mgmt-br and mgmt-bo.

When the network-controller comes up with the changes, the cluster network for mgmt is handled and its removing the vid from "bridge vlan .."

nadtrunk2

nadtrunk1

w13915984028:
@rrajendran17 It is a bug when fetching mgmt vid, I will fix it soon. thanks.

w13915984028:
@Vicente-Cheng The original code can't simply support vlan-trunk feature without massive refactor. It is hard to split the PR atm. thanks.

rrajendran17:
@w13915984028  some observations and questions
1.With the usage of trunk mode, users will have to change the cloud configuration scripts on the VM to include vlan sub interfaces instead of the VM interface.The first interface on the VM generally acquires ip address through dhcp, in this case this will not work, which mandates either to have mgmt (masquerade) interface as first nic to install the necessary packages or have cloud init scripts for VM come up with vlan sub interfaces configurations before start,
This should be made clear to the users via documentation.

2.And looks like kubevirt is not handling propagation of ip address on the  vlan sub interfaces from the guest os to the vmi.I could see the qemu-ga is listing all the vlan sub interfaces with ip addresses correctly but vmi is listing only the last configured vlan sub interface with only pod's ipv6 address and not the configured ipv4 addresses on the interface.We need to check this further.

3.Any reason why the configuration for nad chosen to be min and max range for trunk configuration ?Is it possible to have list of ranges or range + individual vid configuration on the same nad.
Similar to switch port configurations on switches, users will like to have something like 2011-2015,2020,2025 etc.
It is still possible users can do this on another nad, but it will be more flexible if users can specify on the same nad and VM could attach to that NAD.

4.The legacy behavior lets users create two different nads with same vlan id, so a VM can create 2 different  interfaces with same vlan id (same subnet).I think this is not a common scenario required.Do you think we need to restrict this behavior in webhook ? Are there any usecases with this in Harvester ?
If we want to restrict this behaviour, then we should also restrict overlapping range in different nad creation for trunk configuration.

5.Are we disabling "route connectivity check " for trunk vlans nad ? do we have make this optional in future for trunk vlans as well ? As we discussed in another thread,enabled by default and disable by users when required ?

w13915984028:
@rrajendran17 Thanks for those good questions.

> 1.With the usage of trunk mode, users will have to change the cloud configuration scripts on the VM to include vlan sub interfaces instead of the VM interface.The first interface on the VM generally acquires ip address through dhcp, in this case this will not work, which mandates either to have mgmt (masquerade) interface as first nic to install the necessary packages or have cloud init scripts for VM come up with vlan sub interfaces configurations before start, This should be made clear to the users via documentation.

Yes, I will add a KB and update the doc as well.

Here are some user-stories about the `vlan trunk`, the trunk mode nic on guest VM is mainly used as L2 device.

https://github.com/w13915984028/harvester-develop-summary/blob/main/test-harvester-vlan-trunk-mode-nad.md#harvester-on-harvester

On Harvester-on-Harvester, the installer is responsible to setup the mgmt IP;  
on other scenarios, a `tag-nad` is used as the first nic for setting up `management ip` as usual and then the another `trunk-nad` is for L2 usage.

It does not have to rely on mgmt (masquerade) interface.


> 
> 2.And looks like kubevirt is not handling propagation of ip address on the vlan sub interfaces from the guest os to the vmi.I could see the qemu-ga is listing all the vlan sub interfaces with ip addresses correctly but vmi is listing only the last configured vlan sub interface with only pod's ipv6 address and not the configured ipv4 addresses on the interface.We need to check this further.
>

See above.

> 3.Any reason why the configuration for nad chosen to be min and max range for trunk configuration ?Is it possible to have list of ranges or range + individual vid configuration on the same nad. Similar to switch port configurations on switches, users will like to have something like 2011-2015,2020,2025 etc. It is still possible users can do this on another nad, but it will be more flexible if users can specify on the same nad and VM could attach to that NAD.
> 

It is a list with flexible multi inputs like `"vlanTrunk":[{"minID":66,"maxID":88}, {"minID":100,"maxID":200}, {"vid":300}]`, which allows both range, single, and overlapping between them.


> 4.The legacy behavior lets users create two different nads with same vlan id, so a VM can create 2 different interfaces with same vlan id (same subnet).I think this is not a common scenario required.Do you think we need to restrict this behavior in webhook ? Are there any usecases with this in Harvester ? If we want to restrict this behaviour, then we should also restrict overlapping range in different nad creation for trunk configuration.

An interesting scenario is that the `storage-network` and `migration-network` can share one vid with splitting sub-net ranges.
Refer: https://github.com/harvester/harvester/pull/8546

Meanwhile, nad-vid-100 , nad-trunk-50-300, nad-trunk-100-200 are not conflicting, that depends on user's network planning, there is no strong reason to forcely limit the overlapping between nads. Instead, user is free to set the allowed vids for each NAD.

> 5.Are we disabling "route connectivity check " for trunk vlans nad ? do we have make this optional in future for trunk vlans as well ? As we discussed in another thread,enabled by default and disable by users when required ?

At the view of "route connectivity check ", trunk nad has vid 0, similar to un-tag nad. Considering the user-stroies, I prefer to leave it as is for now, no check, always up.

https://github.com/harvester/network-controller-harvester/blob/509931e29652f5f0b10418dde7409cfc85098b3e/pkg/utils/nad.go#L128

On v170, we can refactor the "route connectivity check" with options to enable/disable and more.

rrajendran17:
> @rrajendran17 Thanks for those good questions.
> 
> > 1.With the usage of trunk mode, users will have to change the cloud configuration scripts on the VM to include vlan sub interfaces instead of the VM interface.The first interface on the VM generally acquires ip address through dhcp, in this case this will not work, which mandates either to have mgmt (masquerade) interface as first nic to install the necessary packages or have cloud init scripts for VM come up with vlan sub interfaces configurations before start, This should be made clear to the users via documentation.
> 
> Yes, I will add a KB and update the doc as well.
> 
> Here are some user-stories about the `vlan trunk`, the trunk mode nic on guest VM is mainly used as L2 device.
> 
> https://github.com/w13915984028/harvester-develop-summary/blob/main/test-harvester-vlan-trunk-mode-nad.md#harvester-on-harvester
> 
> On Harvester-on-Harvester, the installer is responsible to setup the mgmt IP; and the rest are used as bridge uplink, no IP;
> 
> on other scenarios, a `tag-nad` is used as the first nic for setting up `management ip` as usual and then the another `trunk-nad` nic is for L2 usage.
> 
> It does not have to rely on mgmt (masquerade) interface.
> 
> > 2.And looks like kubevirt is not handling propagation of ip address on the vlan sub interfaces from the guest os to the vmi.I could see the qemu-ga is listing all the vlan sub interfaces with ip addresses correctly but vmi is listing only the last configured vlan sub interface with only pod's ipv6 address and not the configured ipv4 addresses on the interface.We need to check this further.
> 
> See above.
> 
> > 3.Any reason why the configuration for nad chosen to be min and max range for trunk configuration ?Is it possible to have list of ranges or range + individual vid configuration on the same nad. Similar to switch port configurations on switches, users will like to have something like 2011-2015,2020,2025 etc. It is still possible users can do this on another nad, but it will be more flexible if users can specify on the same nad and VM could attach to that NAD.
> 
> It is a list with flexible multi inputs like `"vlanTrunk":[{"minID":66,"maxID":88}, {"minID":100,"maxID":200}, {"vid":300}]`, which allows both range, single, and overlapping between them.
> 
> > 4.The legacy behavior lets users create two different nads with same vlan id, so a VM can create 2 different interfaces with same vlan id (same subnet).I think this is not a common scenario required.Do you think we need to restrict this behavior in webhook ? Are there any usecases with this in Harvester ? If we want to restrict this behaviour, then we should also restrict overlapping range in different nad creation for trunk configuration.
> 
> An interesting scenario is that the `storage-network` and `migration-network` can share one vid with splitting sub-net ranges. Refer: [harvester/harvester#8546](https://github.com/harvester/harvester/pull/8546)
> 
> Meanwhile, nad-vid-100 , nad-trunk-50-300, nad-trunk-100-200 are not conflicting, that depends on user's network planning, there is no strong reason to forcely limit the overlapping between nads. Instead, user is free to set the allowed vids for each NAD.
> 
> > 5.Are we disabling "route connectivity check " for trunk vlans nad ? do we have make this optional in future for trunk vlans as well ? As we discussed in another thread,enabled by default and disable by users when required ?
> 
> At the view of "route connectivity check ", trunk nad has vid 0, similar to un-tag nad. Considering the user-stroies, I prefer to leave it as is for now, no check, always up.
> 
> https://github.com/harvester/network-controller-harvester/blob/509931e29652f5f0b10418dde7409cfc85098b3e/pkg/utils/nad.go#L128
> 
> On v170, we can refactor the "route connectivity check" with options to enable/disable and more.

Thanks @w13915984028 for the details

1.In the past we have an issue reported from some user where failure to update an ip address in the vmi of a downstream cluster caused provisioning issues.It was a scenario where a VM was created as downstream cluster and it did not have an ipv4 address in the VM instance.Got you, I think we are good since the use case for this is Harvester in Harvester mainly and the mgmt ip configured by installer and connectivity is established, and mainly used for L2 purpose for VMs inside them.
May be if user wants to use this for L3 on the vlan sub interfaces on standalone Harvester VM, then user should be aware that there is a known issue with vlan sub interface on VMs where ips will not be updated in the vmi.
For VM

rrajendran17:
> > @rrajendran17 Thanks for those good questions.
> > > 1.With the usage of trunk mode, users will have to change the cloud configuration scripts on the VM to include vlan sub interfaces instead of the VM interface.The first interface on the VM generally acquires ip address through dhcp, in this case this will not work, which mandates either to have mgmt (masquerade) interface as first nic to install the necessary packages or have cloud init scripts for VM come up with vlan sub interfaces configurations before start, This should be made clear to the users via documentation.
> > 
> > 
> > Yes, I will add a KB and update the doc as well.
> > Here are some user-stories about the `vlan trunk`, the trunk mode nic on guest VM is mainly used as L2 device.
> > https://github.com/w13915984028/harvester-develop-summary/blob/main/test-harvester-vlan-trunk-mode-nad.md#harvester-on-harvester
> > On Harvester-on-Harvester, the installer is responsible to setup the mgmt IP; and the rest are used as bridge uplink, no IP;
> > on other scenarios, a `tag-nad` is used as the first nic for setting up `management ip` as usual and then the another `trunk-nad` nic is for L2 usage.
> > It does not have to rely on mgmt (masquerade) interface.
> > > 2.And looks like kubevirt is not handling propagation of ip address on the vlan sub interfaces from the guest os to the vmi.I could see the qemu-ga is listing all the vlan sub interfaces with ip addresses correctly but vmi is listing only the last configured vlan sub interface with only pod's ipv6 address and not the configured ipv4 addresses on the interface.We need to check this further.
> > 
> > 
> > See above.
> > > 3.Any reason why the configuration for nad chosen to be min and max range for trunk configuration ?Is it possible to have list of ranges or range + individual vid configuration on the same nad. Similar to switch port configurations on switches, users will like to have something like 2011-2015,2020,2025 etc. It is still possible users can do this on another nad, but it will be more flexible if users can specify on the same nad and VM could attach to that NAD.
> > 
> > 
> > It is a list with flexible multi inputs like `"vlanTrunk":[{"minID":66,"maxID":88}, {"minID":100,"maxID":200}, {"vid":300}]`, which allows both range, single, and overlapping between them.
> > > 4.The legacy behavior lets users create two different nads with same vlan id, so a VM can create 2 different interfaces with same vlan id (same subnet).I think this is not a common scenario required.Do you think we need to restrict this behavior in webhook ? Are there any usecases with this in Harvester ? If we want to restrict this behaviour, then we should also restrict overlapping range in different nad creation for trunk configuration.
> > 
> > 
> > An interesting scenario is that the `storage-network` and `migration-network` can share one vid with splitting sub-net ranges. Refer: [harvester/harvester#8546](https://github.com/harvester/harvester/pull/8546)
> > Meanwhile, nad-vid-100 , nad-trunk-50-300, nad-trunk-100-200 are not conflicting, that depends on user's network planning, there is no strong reason to forcely limit the overlapping between nads. Instead, user is free to set the allowed vids for each NAD.
> > > 5.Are we disabling "route connectivity check " for trunk vlans nad ? do we have make this optional in future for trunk vlans as well ? As we discussed in another thread,enabled by default and disable by users when required ?
> > 
> > 
> > At the view of "route connectivity check ", trunk nad has vid 0, similar to un-tag nad. Considering the user-stroies, I prefer to leave it as is for now, no check, always up.
> > https://github.com/harvester/network-controller-harvester/blob/509931e29652f5f0b10418dde7409cfc85098b3e/pkg/utils/nad.go#L128
> > 
> > On v170, we can refactor the "route connectivity check" with options to enable/disable and more.

Got you, Thanks @w13915984028 for the details.I have some follow up questions.

1.In the past we had an issue reported from a user where failure to update an ip address in the vmi of a downstream cluster caused provisioning issues.It was a scenario where a VM was created as downstream cluster and it did not have an ipv4 address in the VM instance.My previous question was due to this.Is it always made sure that downstream cluster VM will have a nodeip when using a trunk nad ? Does this always happen through installer which takes care of mgmt ip configuration or can it happen through VM's first nic ip as node ip ?
FYR - https://github.com/harvester/harvester/issues/7520#issuecomment-2781189075

2.And regarding creation of two different nads with same vid, as you pointed out the conclusion is, it is not best practice but we do not want to restrict it by webhook while still users can use it that way.
https://github.com/harvester/harvester/pull/8546#issuecomment-3019033903
In case of storage networks and Migration networks, I think user provide the range and nad is created inside harvester code and whereabouts allocate the ips from the range provided.But in case of nads created for VM interfaces, two VM networks with the same vid will not have purpose especially when they used as L3 interfaces inside VM host.

starbops (commented):
Thanks for the effort! I left some comments, PTAL.

NOTE: I'm still reviewing and just wanted to let you know some of the results before I finish.

pkg/controller/agent/clusternetwork/controller.go
```diff
@@ -0,0 +1,124 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	/*
+		addedList := added.ToLocalAreasWitoutCIDR()
+		removedList := removed.ToLocalAreasWitoutCIDR()
+	*/
+	addedList := added.ToLocalAreas()
+	removedList := removed.ToLocalAreas()
```
rrajendran17, line 105: I see you have already defined ToLocalAreasWitoutCIDR, since we are not using the cidr for ensuring route connectivity(EnsureRouteViaGateway) in current design, do we still need to store this in LocalArea struct ? can we use ToLocalAreasWitoutCIDR() instead ?

pkg/utils/nad.go
```diff
@@ -88,6 +94,172 @@ func NewLayer3NetworkConf(conf string) (*Layer3NetworkConf, error) {
...
+	}
+	annotations[KeyNetworkRoute] = newRouteAnnotation
+
+	return annotations, nil
+}
+
+// caller ensurs this function is only called when oldNad and newNad have same vlan id
+// the first return is not nil only when real change happens
+func OutdateNadLayer3NetworkConfWhenRouteModeChanges(oldNad, newNad *nadv1.NetworkAttachmentDefinition) (map[string]string, error) {
+	if newNad == nil || oldNad == nil {
```
ibrokethecloud, line 152: is this scenario even possible? 

The only call to this method is during an update of the net-attach-def object

pkg/webhook/nad/mutator.go
```diff
@@ -94,18 +88,19 @@ func (m *Mutator) Resource() admission.Resource {
...
 
-func (m *Mutator) ensureLabels(nad *cniv1.NetworkAttachmentDefinition, oldConf, newConf *utils.NetConf) (admission.Patch, error) {
+func (m *Mutator) ensureLabels(nad *cniv1.NetworkAttachmentDefinition, _, newConf *utils.NetConf) (admission.Patch, error) {
 	labels := nad.Labels
 	if labels == nil {
 		labels = make(map[string]string)
 	}
 
-	if newConf.Type == utils.CNITypeKubeOVN {
+	if newConf.IsKubeOVNCNI() {
```
ibrokethecloud, line 97: @rrajendran17 do we allow a user to reconfigure an existing nad from vlan to kubeovn type? 

I think we should reject such requests and force user to recreate the NAD using the correct type.

pkg/controller/manager/nad/controller.go
```diff
@@ -148,27 +150,42 @@ func (h Handler) OnCNChange(_ string, cn *networkv1.ClusterNetwork) (*networkv1.
...
+		if err := h.EnsureJob2GetLayer3NetworkInfo(nad, netconf); err != nil {
+			return nil, err
+		}
 	}
-
-	klog.Infof("nad configuration %s has been changed: %s", nad.Name, nad.Spec.Config)
-
-	if err := h.EnsureJob2GetLayer3NetworkInfo(nad); err != nil {
+	// nad change triggers the re-compute of cn's vlanset
+	if err := h.UpdateClusetNetworkVlanSet(nad); err != nil {
```
ibrokethecloud, line 188: would it not be better to just have a relatedresource watch in the cluster controller to watch NAD objects and trigger reconcile of cluster networks when a NAD changes? this would make it easier to follow the code.

pkg/network/vlan/vlan.go
```diff
@@ -59,6 +55,14 @@ func GetVlan(name string) (*Vlan, error) {
 	return v, nil
 }
 
+// by default, each clusternetwork (L2 type) is mapped to a bridge
+// when one node is not selected on the clusternetwork, then it does not have the bridge
+// mgmt is populated to all nodes
+func IsClusterNetworkBridgeSetup(cnName string) bool {
```
ibrokethecloud, line 61: seems unused

pkg/webhook/nad/mutator.go
```diff
@@ -114,27 +109,15 @@ func (m *Mutator) ensureLabels(nad *cniv1.NetworkAttachmentDefinition, oldConf,
...
+	if err != nil {
+		return nil, err
 	}
 
-	cnName := newConf.BrName[:len(newConf.BrName)-len(iface.BridgeSuffix)]
-	labels[utils.KeyClusterNetworkLabel] = cnName
-	if oldConf.BrName != newConf.BrName {
-		labels[utils.KeyLastClusterNetworkLabel] = oldConf.BrName[:len(oldConf.BrName)-len(iface.BridgeSuffix)]
+	// check and get the bridge name
+	cnName, err := utils.GetBridgeNamePrefix(newConf.BrName)
```
ibrokethecloud, line 118: nit: we already set cn name to a label during SetNetworkInfoToLabels.

```
lbs[KeyClusterNetworkLabel] = cnname
```

We can just look this from the map and simplify logic

pkg/webhook/nad/mutator.go
```diff
@@ -152,41 +135,43 @@ func (m *Mutator) ensureLabels(nad *cniv1.NetworkAttachmentDefinition, oldConf,
...
+func tagRouteOutdated(oldNad, newNad *cniv1.NetworkAttachmentDefinition, oldConf, newConf *utils.NetConf) (admission.Patch, error) {
+	if newConf.IsKubeOVNCNI() {
 		return nil, nil
 	}
-
-	annotations := nad.Annotations
-	if annotations == nil {
-		annotations = make(map[string]string)
+	// even when vlan is not changed, the route mode can be changed from `auto` to `static` or vice versa
+	if newNad.Annotations == nil || len(newNad.Annotations) == 0 {
```
ibrokethecloud, line 144: i assume this can only happen when a NAD has been created but controller has not yet discovered the route from completed job?

pkg/webhook/nad/mutator.go
rrajendran17: @ibrokethecloud No users are not allowed change between bridge and ovn Network types.We restrict it in the nad webhook validator.

pkg/controller/agent/clusternetwork/controller.go
w13915984028: original code does not fully decided to remove CIDR, now it is clear, I will update the code

pkg/controller/agent/vlanconfig/controller.go
```diff
@@ -178,31 +179,26 @@ func matchClusterNetwork(vc *networkv1.VlanConfig, vs *networkv1.VlanStatus) boo
...
 	if setupErr != nil {
 		goto updateStatus
 	}
-	// set up VLAN
+	// set up VLAN bridge
 	v = vlan.NewVlan(vc.Spec.ClusterNetwork)
 	if setupErr = v.Setup(uplink); setupErr != nil {
 		goto updateStatus
 	}
-	setupErr = h.AddLocalAreas(v, localAreas)
```
ibrokethecloud, line 198: earlier if had error setting up bridge interface we would skip the setup of local network using `goto updateStatus`

now that this is not needed we could just avoid the entire usage of jump/label in the flow

pkg/controller/agent/clusternetwork/controller.go
```diff
@@ -0,0 +1,107 @@
...
+
+	err = v.AddLocalAreas(added)
+	if err != nil {
+		return nil, err
+	}
+	err = v.RemoveLocalAreas(removed)
+	if err != nil {
+		return nil, err
+	}
+
```
mingshuoqiu: If AddLocalAreas succeeds but RemoveLocalAreas fails, the system is left in a partially updated state.

Should be better with the following
if err := v.AddLocalAreas(added); err != nil {
    return nil, fmt.Errorf("add VLANs failed: %w", err)
}
if err := v.RemoveLocalAreas(removed); err != nil {
    klog.Warningf("failed to remove VLANs: %v", err)
}

pkg/network/iface/link.go
```diff
@@ -53,7 +54,7 @@ func (l *Link) AddBridgeVlan(vid uint16) error {
 // DelBridgeVlan adds a new vlan filter entry
```
mingshuoqiu: Not really about the code change.
But this comment should be `// DelBridgeVlan deletes a vlan filter entry`

pkg/controller/agent/clusternetwork/controller.go
```diff
@@ -0,0 +1,107 @@
...
+	cnVlans, err := utils.GeVlanIDSetFromClusterNetwork(cn.Name, h.nadCache)
+	if err != nil {
+		klog.Infof("cluster network %s failed to get vlanset %s", cn.Name, err.Error())
+		return nil, err
+	}
+
+	// user might configure vlan sub-interface on the existing bridge for additional usage
+	// those vids are out of any nads
+	// try to keep them as more as possbile, and log errors if failed to list or add
+	manualVlans, err := iface.GetManuallyConfiguredVlans(cn.Name)
```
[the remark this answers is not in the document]
w13915984028: The current code, is fetching any device like `cn2-br.2025`, not from `cn2-bo.2025`, anything wrong here?

https://github.com/harvester/network-controller-harvester/blob/61fab3a8e5b9090dd88cd1a89e9b10256ad4fba7/pkg/network/iface/link.go#L237

https://github.com/harvester/network-controller-harvester/blob/61fab3a8e5b9090dd88cd1a89e9b10256ad4fba7/pkg/utils/bridge.go#L28

PR #193, opened 2025-07-15
Run mutator and validator all the time when change happens

w13915984028:
**Problem:**

Observed when testing https://github.com/harvester/network-controller-harvester/pull/186

**Solution:**

Cherry-pick some changes to merge first.

**Related Issue:**

**Test plan:**

Create vlanconfig, check it's labes are always updated.


w13915984028:
@mergify backport v1.6

w13915984028:
Still wait https://github.com/harvester/network-controller-harvester/pull/186 to merge together

PR #211, opened 2025-09-09
Fix route add issues during VM Netwotk addition

rrajendran17:
**Problem:**
Adding a new route of the CIDR of newly created VM Vlan network will return error due to a nil gateway.
https://github.com/harvester/harvester/issues/8910#issuecomment-3222023316

**Solution:**
Remove the validation for gateway exists for every CIDR of the newly created VM as a default route exists via the mgmt interface on every Harvester node in the cluster.

**Related Issue:**
https://github.com/harvester/harvester/issues/8910

**Test plan:**
1.Have a Harvester node with mgmt interface on a vlan. (example vlan2021)
2.Create a cluster network and add network config and wait for network config to be ready status
3.Create a VM Network in vlan 2021 under the cluster network created in step 2.
4.Reboot the Node
5.Check for the status of network config under the cluster network created in step 2, after the node comes up.

Expected behaviour:
The status of the network config under the cluster network created in step 2 should be "ready"

Actual Behavior:
The status of the network config under the cluster network created in step 2 is "not ready"

Slack channel discussions on why EnsureRouteViaGateway, DeleteRouteViaGateway function can be removed:
https://suse.slack.com/archives/C03E4UF4GVD/p1752504466671259

For now, retaining the definitions of the functions.


starbops (approved):
LGTM, thank you!

Vicente-Cheng:
@mergifyio backport v1.6

PR #256, opened 2026-04-01
validate vlanconfig cluster network label

rrajendran17:
**Problem:**
Updating an invalid or non existent cluster network to the label of vlanconfig is allowed today.

**Solution:**
1.User updating the label not same as cluster network name must be blocked.

**Related Issue:**
https://github.com/harvester/harvester/issues/9332

**Test plan:**

1.Create a vlanconfig with cluster network cluster-1
2.Update the label cluster network to "cluster-2" by editing the vlanconfig.

webhook error:update fails as label for cluster network does not match the cluster network name in spec

### DIFF
--- a/pkg/webhook/vlanconfig/validator.go
+++ b/pkg/webhook/vlanconfig/validator.go
@@ -8,6 +8,7 @@ import (
 
 	mapset "github.com/deckarep/golang-set/v2"
 	"github.com/harvester/webhook/pkg/server/admission"
+	ctlcorev1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/core/v1"
 	admissionregv1 "k8s.io/api/admissionregistration/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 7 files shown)


pkg/webhook/vlanconfig/validator.go
```diff
@@ -57,15 +57,19 @@ var _ admission.Validator = &Validator{}
...
-	if _, err := utils.IsClusterNetworkNameValid(vc.Spec.ClusterNetwork); err != nil {
-		return fmt.Errorf(createErr, vc.Name, err)
-	}
-
 	if vc.Spec.ClusterNetwork == utils.ManagementClusterNetworkName {
 		return fmt.Errorf(createErr, vc.Name, fmt.Errorf("cluster network can't be %s",
 			utils.ManagementClusterNetworkName))
 	}
 
+	if valid, label := utils.IsClusterNetworkLabelValid(vc); !valid {
```
w13915984028: just a question, this might cause existing invalid instances be broken, could we update the mutator to ensure the label is always auto amended to parent clusternetwork? thanks.

https://github.com/harvester/network-controller-harvester/blob/a1e725d1a7327557a2984d29511c6b4660fb3817/pkg/webhook/vlanconfig/mutator.go#L43

rrajendran17: vlanconfigs with invalid existing cluster network labels will be non-functional.User has to delete them or update them matching correct cluster network so that the validation pass.sorry, I am not sure if I understand your question correctly. Are you pointing to any particular user scenario ?

w13915984028: a vlanconfig change triggers mutator and validator, if we use mutator to amend the label, then it should be always matching the clusternetwork

rrajendran17: https://github.com/harvester/network-controller-harvester/blob/a1e725d1a7327557a2984d29511c6b4660fb3817/pkg/webhook/vlanconfig/mutator.go#L54 during update only if the cluster network name changes, we amend the label from cluster network name.

w13915984028: could we change the mutator to ensure the consistency all the time?

rrajendran17: Thats doable, but I am not sure if that will be the right approach.The user tries to change something which they are not supposed to do and are not aware that its non-configurable if backend just updates to right cluster network name in label.

rrajendran17:
@w13915984028 The validator now takes care of empty label check also.Thanks for the review.

w13915984028 (dismissed):
LGTM, thanks.

w13915984028 (approved):
LGTM, thanks.

PR #259, opened 2026-04-23
Extend VlanConfig protection to cover RWX network NAD

brandboat:
**Problem:**

**Solution:**

**Related Issue:**

https://github.com/harvester/harvester/issues/10235

**Test plan:**

### TC-01: VlanConfig update is blocked when RWX network NAD exists
Steps:
1. Configure RWX network setting on cluster network cn-test, confirm rwx-network-* NAD appears in harvester-system
1. Attempt to update the VlanConfig on cn-test (e.g., change MTU)
Expected: Rejected with error containing rwx network

### TC-02: VlanConfig delete is blocked when RWX network NAD exists
Steps:
1. Same setup as TC-01
1. Attempt to delete the VlanConfig
Expected: Rejected with error containing rwx network

### TC-03: VlanConfig update/delete is allowed after RWX network NAD is removed
Steps:
1. Remove the RWX network setting (NAD deleted from harvester-system)
1. Attempt to update then delete the VlanConfig
Expected: Both succeed

### TC-04: VlanConfig on a different cluster network is not affected
Steps:
1. RWX NAD exists on cn-test
2. Attempt to update/delete a VlanConfig on a different cluster network
Expected: Succeeds without error

### TC-05: RWX NAD itself cannot be directly modified
Steps:
- Attempt to kubectl patch the rwx-network-* NAD spec
Expected: Rejected with error containing rwxnetwork

### TC-06: Regression tests for storage-network
Step: follow TC-01~TC-05, but create storage-network instead.

### DIFF
--- a/pkg/utils/nad.go
+++ b/pkg/utils/nad.go
@@ -629,32 +629,28 @@ func isMaskZero(ipnet *net.IPNet) bool {
 	return true
 }
 
-// if this nad is a storagenetwork nad
-func IsStorageNetworkNad(nad *nadv1.NetworkAttachmentDefinition) bool {
+// IsSystemNetworkNad checks if the nad is a system network nad (storage or rwx).
+func IsSystemNetworkNad(nad *nadv1.NetworkAttachmentDefinition) bool {
 	if nad == nil || nad.Namespace != HarvesterSystemNamespaceName {
 		return false
 	}
 
-	// seems Harvester webhook has no protection on this annotation
-	if nad.Annotations != nil && nad.Annotations[StorageNetworkAnnotation] == "true" {
-		return true
+	isMatch := func(annotation, prefix string) bool {
+		if nad.Annotations != nil && nad.Annotations[annotation] == "true" {
+			return true
+		}
+		return strings.HasPrefix(nad.Name, prefix)
 	}
 
-	// check name
-	if strings.HasPrefix(nad.Name, StorageNetworkNetAttachDefPrefix) {
-		return true
-	}
-
-	return false
+	return isMatch(StorageNetworkAnnotation, StorageNetworkNetAttachDefPrefix) ||
+		isMatch(RWXNetworkAnnotation, RWXNetworkNetAttachDefPrefix)
 }
 
-// filter the first active storage network nad from a list of nads
-func FilterFirstActiveStorageNetworkNad(nads []*nadv1.NetworkAttachmentDefinition) *nadv1.NetworkAttachmentDefinition {
-	if len(nads) == 0 {
-		return nil
-	}
+// FilterFirstActiveSystemNetworkNad checks for any active system network nad (storage or rwx) from a list of nads.
+// Returns the nad and its type description, or nil if none found.
+func FilterFirstActiveSystemNetworkNad(nads []*nadv1.NetworkAttachmentDefinition) *nadv1.NetworkAttachmentDefinition {
 	for _, nad := range nads {
-		if IsStorageNetworkNad(nad) && nad.DeletionTimestamp == nil {
+		if nad != nil && nad.DeletionTimestamp == nil && IsSystemNetworkNad(nad) {
 			return nad
 		}
 	}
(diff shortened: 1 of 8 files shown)


w13915984028 (dismissed):
LGTM, thanks.

pkg/webhook/vlanconfig/validator.go
```diff
@@ -278,22 +287,30 @@ func (v *Validator) validateMTU(current *networkv1.VlanConfig) error {
...
+
+// if storagenetwork nad is there, and affected node number > 0, then deny
+func (v *Validator) checkStorageNetwork(vc *networkv1.VlanConfig, nodes mapset.Set[string]) error {
+	return v.checkNetworkNadAttached(vc, nodes, (*utils.NadGetter).GetFirstActiveStorageNetworkNadOnClusterNetwork, "storage network")
+}
+
+// if rwx network nad is there, and affected node number > 0, then deny
+func (v *Validator) checkRWXNetwork(vc *networkv1.VlanConfig, nodes mapset.Set[string]) error {
+	return v.checkNetworkNadAttached(vc, nodes, (*utils.NadGetter).GetFirstActiveRWXNetworkNadOnClusterNetwork, "rwx network")
+}
```
rrajendran17: Functionally looks good to me, but can we simplify this ? Example have a common single function from Update/Delete vlanconfig and compare for either of the annotations to be true and corresponding error string in return.
As we have set of nads from same namespace for both just extending it to include RWX annotation and rwx prefix name should do.WDYT ?

pkg/webhook/vlanconfig/validator_test.go
```diff
@@ -510,6 +510,92 @@ func TestUpdateVlanConfig(t *testing.T) {
...
+				},
+			},
+			currentNAD: &cniv1.NetworkAttachmentDefinition{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "rwx-network-abc123",
+					Namespace: utils.HarvesterSystemNamespaceName,
+					Labels:    map[string]string{utils.KeyClusterNetworkLabel: testCnName},
+				},
+			},
+		},
```
rrajendran17, line 599: can you also include test case to check for storage network annotation and rwx annotation?

pkg/utils/nad.go
```diff
@@ -629,38 +629,45 @@ func isMaskZero(ipnet *net.IPNet) bool {
...
+
+// if this nad is a rwx network nad
+func IsRWXNetworkNad(nad *nadv1.NetworkAttachmentDefinition) bool {
+	return isSystemNetworkNad(nad, RWXNetworkAnnotation, RWXNetworkNetAttachDefPrefix)
+}
+
+// filter the first active rwx network nad from a list of nads
+func FilterFirstActiveRWXNetworkNad(nads []*nadv1.NetworkAttachmentDefinition) *nadv1.NetworkAttachmentDefinition {
+	return filterFirstActiveNad(nads, IsRWXNetworkNad)
+}
```
rrajendran17: https://github.com/harvester/network-controller-harvester/blob/a1e725d1a7327557a2984d29511c6b4660fb3817/pkg/webhook/clusternetwork/validator.go#L215 this checks if storage network is active during mgmt network update from cli for MTU, similarly it should check for rwx network as well.

It will be better if we have common function so that we do not miss cases for RWX.

pkg/webhook/vlanconfig/validator_test.go
brandboat: Thanks, I added the rwx|storage-network NAD validation test to TestIsSystemNetworkNad. Otherwise we have to  duplicate the same test case across all validators, for me it's a bit redundant.

pkg/utils/nad.go
brandboat: Thanks, all comments are addressed in the latest commit

pkg/utils/nad.go
```diff
@@ -629,32 +629,28 @@ func isMaskZero(ipnet *net.IPNet) bool {
 	return true
 }
 
-// if this nad is a storagenetwork nad
-func IsStorageNetworkNad(nad *nadv1.NetworkAttachmentDefinition) bool {
+// IsSystemNetworkNad checks if the nad is a system network nad (storage or rwx).
+func IsSystemNetworkNad(nad *nadv1.NetworkAttachmentDefinition) bool {
 	if nad == nil || nad.Namespace != HarvesterSystemNamespaceName {
```
rrajendran17, line 634: nad == nil could be redundant.Please check.

rrajendran17:
LGTM, Thanks for taking care of the comments.

rrajendran17 (approved):

pkg/utils/nad.go
brandboat: Yeah, it might seem redundant for now, but I'd prefer to keep this nil check since it's a public function, removing it would make future callers that pass nil panic.

PR #260, opened 2026-04-24
skip vlanconfig with empty matched nodes

rrajendran17:
Problem:
hostnetworkconfig cannot be created if there exists vlanconfig with empty matched nodes.

Solution:
Allow creation of hostnetworkconfig if it has vlanconfigs with empty matched nodes if there are other vlanconfigs present for the same cluster spanning all nodes.

Related Issue(s):

https://github.com/harvester/harvester/issues/9332

Test plan:

1.Create a cluster network "cluster"
2.Create a vlanconfig1 under "cluster" as with some random node selector (select the nic, then add node selector)
3.Create another vlanconfig2 under the same cluster "cluster" selecting all nodes in the cluster and uplink physical interface.
4.The UI will show 2 vlanconfigs with vlanconfig1 as "not ready" and vlanconfig2 as ready.
5.Create a hostnetworkconfig using cluster network "cluster"
6.hostnetworkconfig creation must be successful

Without the fix:
The webhook will throw "vlanconfig annotations is absent for matched nodes" while hostnetworkconfig creation as one of the vlanconfig exists with empty matched nodes list.

Additional documentation or context

### DIFF
--- a/pkg/webhook/hostnetworkconfig/validator.go
+++ b/pkg/webhook/hostnetworkconfig/validator.go
@@ -357,11 +357,13 @@ func (v *Validator) checkVlanStatusReady(clusterNetwork string) error {
 }
 
 func getMatchNodes(vc *networkv1.VlanConfig) ([]string, error) {
-	if vc.Annotations == nil || vc.Annotations[utils.KeyMatchedNodes] == "" {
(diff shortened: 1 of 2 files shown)


w13915984028 (approved):
LGTM, thanks.

PR #266, opened 2026-05-14
validate VM Nad exclusive of storage network vlan

rrajendran17:
**Problem:**
VM Network and Storage network can use same vlan currently which do not allow proper isolation of storage network traffic.
By enabling exclusive vlan on storage network, network controller can make sure to reject NAD config with same vlan id as storage network enabling proper isolation

**Solution:**
Reject VM Networks with vid same as storage network if exclusive vlan is configured on storage network settings

**Related Issue:**
https://github.com/harvester/harvester/issues/9622

**Test plan:**
case 1: VM Network access or untagged 
1.Create storage network with vlan id 27 successfully and update the settings to include exclusiveVlan as true as below
```
value: '{"vlan":27,"exclusiveVlan":true,"clusterNetwork":"cluster","range":"27.0.0.0/24"}'
```
2.Create a VM Network L2VlanNetwork with mode access with vlan-id 27
3.It should be rejected by webhook 

case 2: Trunk vlan
1. storage network created from case 1 exists.
2.Create a VM Network L2VlanNetwork with mode trunk with min and max ranges 25-30,100-200,30-40
3.It should be rejected by webhook 

case 3: Create access or trunk vlan which not have vlan 27
The configuration should be accepted and VM Network must be created successfully.


rrajendran17:
Use exclusive vlan annotation in storage network nad for validation https://github.com/harvester/network-controller-harvester/pull/267

PR #267, opened 2026-05-19
validate VM Network vlan exclusive of storage network vlan

rrajendran17:
Problem:
VM Network and Storage network can use same vlan currently which do not allow proper isolation of storage network traffic.
By enabling exclusive vlan on storage network, network controller can make sure to reject NAD config with same vlan id as storage network enabling proper isolation

Solution:
Reject VM Networks with vid same as storage network if exclusive vlan is configured on storage network settings

Related Issue:
https://github.com/harvester/harvester/issues/9622

Test plan:
case 1: VM Network access or untagged
1.Create storage network with vlan id 27 successfully and update the settings to include exclusiveVlan as true as below
value: '{"vlan":27,"exclusiveVlan":true,"clusterNetwork":"cluster","range":"27.0.0.0/24"}'

2.Create a VM Network L2VlanNetwork with mode access with vlan-id 27
3.It should be rejected by webhook

case 2: Trunk vlan
1.storage network created from case 1 exists.
2.Create a VM Network L2VlanNetwork with mode trunk with min and max ranges 25-30,100-200,30-40
3.It should be rejected by webhook

case 3: Create access or trunk vlan which not have vlan 27
The configuration should be accepted and VM Network must be created successfully.

### DIFF
--- a/pkg/webhook/nad/validator.go
+++ b/pkg/webhook/nad/validator.go
@@ -59,6 +59,10 @@ func (v *Validator) Create(_ *admission.Request, newObj runtime.Object) error {
 		return fmt.Errorf(createErr, nad.Namespace, nad.Name, err)
 	}
 
+	if err := v.checkVlanOverlapWithSN(nad, conf); err != nil {
+		return fmt.Errorf(createErr, nad.Namespace, nad.Name, err)
+	}
+
 	//allow overlay nad creation only if subnet crd/kubeovn is enabled
 	if conf.IsKubeOVNCNI() {
 		if !v.subnetEnabled {
@@ -102,6 +106,10 @@ func (v *Validator) Update(_ *admission.Request, oldObj, newObj runtime.Object) 
 	// skip the following check if the config is not changed
 	if reflect.DeepEqual(newConf, oldConf) {
 		return nil
+	}
+
+	if err := v.checkVlanOverlapWithSN(newNad, newConf); err != nil {
+		return fmt.Errorf(updateErr, newNad.Namespace, newNad.Name, err)
 	}
 
 	if err := v.checkNadTypes(oldConf, newConf); err != nil {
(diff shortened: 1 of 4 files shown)


ibrokethecloud (dismissed):

pkg/webhook/nad/validator.go
```diff
@@ -393,6 +401,34 @@ func (v *Validator) checkOverlayVMsUsingClusterNetwork(nad *cniv1.NetworkAttachm
...
+	}).AsSelector())
+	if err != nil {
+		return fmt.Errorf("failed to list storage network nads with exclusive vlan enabled, err %w", err)
+	}
+
+	if len(snNads) == 0 {
+		return nil
+	}
+
+	snNadConf, decodeErr := utils.DecodeNadConfigToNetConf(snNads[0])
```
w13915984028, line 421: if both of SN, RWX  are using exclusive VIDs, should all NADs here are checked?

it seems also essential to avoid check vid 1.

w13915984028 (dismissed):
LGTM, thanks.

pkg/webhook/nad/validator.go
rrajendran17: exclusive vid is supported only for storage network NAD and ExclusiveVlanStorageNetworkLabel will be set only on storage network NAD.(no rwx NAD).

Both VM Network and storage network supports vid range from 1-4094, so its better to exclude vid 1 if exclusive vlan is enabled.In case user wants to use vlan as default vlan and needs both storage network and vm network in vlan 1, then user could disable exclusive vlan and use it.

pkg/webhook/nad/validator.go
```diff
@@ -393,6 +401,34 @@ func (v *Validator) checkOverlayVMsUsingClusterNetwork(nad *cniv1.NetworkAttachm
 	return nil
 }
 
+func (v *Validator) checkVlanOverlapWithSN(nad *cniv1.NetworkAttachmentDefinition, nadConf *utils.NetConf) error {
+	//check only for vlan nad with type bridge CNI
+	if utils.IsSystemNetworkNad(nad) || nadConf.IsKubeOVNCNI() {
+		return nil
```
rrajendran17, line 407: Excluded overlay NAD in the check.BTW, good catch

w13915984028 (approved):
LGTM, thanks.

PR #276, opened 2026-06-04
vlanconfig info for mgmt clusternetwork

rrajendran17:
**Problem:**
Mgmt cluster network does not display the network information in UI.
Users needs the information on the mgmt interface like MTU, interfaces and bond mode which is not available in UI currently like other user defined cluster networks.
Currently the backend does not have network information of the mgmt cluster network.

**Solution:**
Create a vlanconfig per node by the vlanconfig agent controller.
- vlanconfig agent during startup initializes a vlanconfig resource for the mgmt cluster network by retrieving network information like MTU, slaves in the mgmt-bo, and bond mode using ip commands (mgmt-vlanconfig-<nodename>)
- match by mgmt cluster network rule on the link monitor agent controller to reconcile vlanconfig whenever there are changes to link parameters of mgmt-bo/mgmt-br
- Restrict create/update/delete operations on mgmt vlanconfig from user (allow only from network controller)
- skip node updates on mgmt vlanconfig
- skip other vlanconfig operations (like setup/remove vlan/uplink) on mgmt vlanconfigs
- vlanstatus are not created for mgmt vlanconfig as they are not manged by network controller

**Related Issue:**
https://github.com/harvester/harvester/issues/6110

**Test plan:**
case 1:
1.Once the network controller modules come up, check kubectl get vlanconfig - mgmt-vlanconfig must be available on per node.
2.Check if link attributes like MTU,bond mode, slaves in bond mode match the configuration done during installation.

case 2: Add/Remove slave interfaces to the mgmt-bo (any node)
1.sudo ip link add ens7 type dummy
2.sudo ip link set ens7 master mgmt-bo
3.Verify kubectl get vlanconfig mgmt-vlanconfig -o yaml (ens7 added to the links)
4.sudo ip link set ens7 nomaster
5.Verify kubectl get vlanconfig mgmt-vlanconfig-<node-name> -o yaml (ens7 removed from the links)

case 3: Update MTU on mgmt-bo
1.sudo ip link set mgmt-bo mtu 1600
2..Verify kubectl get vlanconfig mgmt-vlanconfig -o yaml (mtu changed to 1600)
3.sudo ip link set mgmt-bo mtu 1500
4..Verify kubectl get vlanconfig mgmt-vlanconfig-<node-name> -o yaml (mtu changed to 1500)

case 4: Restrict create/update/delete of mgmt vlanconfig
1.Create or update or delete the mgmt vlanconfig - webhook will throw error

UI changes:

https://github.com/harvester/harvester-ui-extension/pull/974

### DIFF
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -87,7 +87,7 @@ func main() {
 			Name:        "controller-user",
 			EnvVar:      "CONTROLLER_USER_NAME",
 			Destination: &options.ControllerUsername,
-			Value:       "harvester-load-balancer-webhook",
+			Value:       "system:serviceaccount:harvester-system:harvester-network-controller",
 			Usage:       "The harvester controller username",
 		},
 		cli.StringFlag{
--- a/pkg/controller/agent/linkmonitor/controller.go
+++ b/pkg/controller/agent/linkmonitor/controller.go
@@ -7,12 +7,14 @@ import (
 	ctlcorev1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/core/v1"
 	"github.com/sirupsen/logrus"
 	"github.com/vishvananda/netlink"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
 
 	networkv1 "github.com/harvester/harvester-network-controller/pkg/apis/network.harvesterhci.io/v1beta1"
 	"github.com/harvester/harvester-network-controller/pkg/config"
 	ctlnetworkv1 "github.com/harvester/harvester-network-controller/pkg/generated/controllers/network.harvesterhci.io/v1beta1"
 	"github.com/harvester/harvester-network-controller/pkg/network/monitor"
+	"github.com/harvester/harvester-network-controller/pkg/utils"
 )
 
 const (
@@ -23,6 +25,8 @@ type Handler struct {
 	nodeName string
 
 	nodeCache    ctlcorev1.NodeCache
+	vcCache      ctlnetworkv1.VlanConfigCache
+	vcClient     ctlnetworkv1.VlanConfigClient
 	lmController ctlnetworkv1.LinkMonitorController
 	lmClient     ctlnetworkv1.LinkMonitorClient
 
@@ -32,10 +36,13 @@ type Handler struct {
 func Register(ctx context.Context, management *config.Management) error {
 	lms := management.HarvesterNetworkFactory.Network().V1beta1().LinkMonitor()
 	nodes := management.CoreFactory.Core().V1().Node()
+	vcs := management.HarvesterNetworkFactory.Network().V1beta1().VlanConfig()
 
 	h := &Handler{
 		nodeName:     management.Options.NodeName,
 		nodeCache:    nodes.Cache(),
+		vcCache:      vcs.Cache(),
+		vcClient:     vcs,
 		lmController: lms,
 		lmClient:     lms,
 	}
@@ -75,6 +82,10 @@ func (h Handler) OnChange(_ string, lm *networkv1.LinkMonitor) (*networkv1.LinkM
 
 	if err := h.syncLinkStatus(lm); err != nil {
 		return nil, fmt.Errorf("synchronize links to link monitor %s failed, error: %w", lm.Name, err)
+	}
+
+	if lm.Name == utils.ManagementClusterNetworkName {
+		return nil, h.updateMgmtVlanConfig()
 	}
 
 	return lm, nil
@@ -206,4 +217,36 @@ func (h Handler) syncLinkStatus(lm *networkv1.LinkMonitor) error {
 	}
 
 	return h.updateStatus(lm, linkStatusList)
+}
+
+// update mgmt vlanconfig when mgmt cluster link params change
+func (h Handler) updateMgmtVlanConfig() error {
+	mgmtVlanConfigName := utils.GetMgmtVlanConfigName(h.nodeName)
+	vc, err := h.vcCache.Get(mgmtVlanConfigName)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			logrus.Infof("no VlanConfig found for mgmt cluster network")
+			return nil
+		}
+		return fmt.Errorf("failed to get VlanConfig %s: %w", mgmtVlanConfigName, err)
+	}
+
+	mgmtIntfInfo, err := utils.GetMgmtBondInfo()
+	if err != nil {
+		logrus.Errorf("failed to get mgmt interface info: %v", err)
+		return err
+	}
+	vcCopy := vc.DeepCopy()
+
+	if !utils.BuildUpdatedVlanConfig(vcCopy, mgmtIntfInfo) {
+		return nil
+	}
+
+	_, err = h.vcClient.Update(vcCopy)
+	if err != nil {
+		return fmt.Errorf("failed to update VlanConfig %s: %w", vcCopy.Name, err)
+	}
+
+	logrus.Infof("updated VlanConfig %s with new info: MTU=%d, BondMode=%s, NICs=%v", vcCopy.Name, vcCopy.Spec.Uplink.LinkAttrs.MTU, vcCopy.Spec.Uplink.BondOptions.Mode, vcCopy.Spec.Uplink.NICs)
+	return nil
 }
--- a/pkg/controller/agent/vlanconfig/controller.go
+++ b/pkg/controller/agent/vlanconfig/controller.go
@@ -79,9 +79,10 @@ func Register(ctx context.Context, management *config.Management) error {
 }
 
 func (h Handler) OnChange(_ string, vc *networkv1.VlanConfig) (*networkv1.VlanConfig, error) {
-	if vc == nil || vc.DeletionTimestamp != nil {
+	if vc == nil || vc.Spec.ClusterNetwork == utils.ManagementClusterNetworkName || vc.DeletionTimestamp != nil {
 		return nil, nil
 	}
+
 	logrus.Infof("vlan config %s has been changed, spec: %+v", vc.Name, vc.Spec)
 
 	isMatched, err := h.MatchNode(vc)
@@ -113,7 +114,7 @@ func (h Handler) OnChange(_ string, vc *networkv1.VlanConfig) (*networkv1.VlanCo
 }
 
 func (h Handler) OnRemove(_ string, vc *networkv1.VlanConfig) (*networkv1.VlanConfig, error) {
-	if vc == nil {
+	if vc == nil || vc.Spec.ClusterNetwork == utils.ManagementClusterNetworkName {
 		return nil, nil
 	}
 
@@ -137,6 +138,11 @@ func (h Handler) initialize() error {
 	if err := iface.DisableBridgeNF(); err != nil {
 		return fmt.Errorf("disable net.bridge.bridge-nf-call-iptables failed, error: %v", err)
 	}
+
+	if err := h.CreateOrUpdateMgmtVlanConfig(); err != nil {
+		return fmt.Errorf("mgmt vlanconfig error: %w", err)
+	}
+
 	return nil
 }
 
@@ -461,4 +467,89 @@ func (h Handler) removeNodeLabel(vs *networkv1.VlanStatus) error {
 // vlanstatus name: <vc name>-<node name>-<crc32 checksum>
 func (h Handler) statusName(clusterNetwork string) string {
 	return utils.Name("", clusterNetwork, h.nodeName)
+}
+
+func (h *Handler) createMgmtVlanConfig(vc *networkv1.VlanConfig) error {
+	_, err := h.vcClient.Create(vc)
+	if err != nil {
+		if apierrors.IsAlreadyExists(err) {
+			logrus.Debugf("%s already exists", utils.GetMgmtVlanConfigName(h.nodeName))
+			return nil
+		}
+
+		return fmt.Errorf("failed to create mgmt-vlanconfig %s: %w",
+			utils.GetMgmtVlanConfigName(h.nodeName), err)
+	}
+
+	logrus.Infof("created VlanConfig %s with MTU=%d BondMode=%s NICs=%v", vc.Name, vc.Spec.Uplink.LinkAttrs.MTU, vc.Spec.Uplink.BondOptions.Mode, vc.Spec.Uplink.NICs)
+	return nil
+}
+
+// CreateOrUpdateMgmtVlanConfig checks mgmt-bo, gathers info, and creates or updates a VlanConfig resource.
+func (h *Handler) CreateOrUpdateMgmtVlanConfig() error {
+	mgmtInfo, err := utils.GetMgmtBondInfo()
+	if err != nil {
+		logrus.Warnf("mgmt interface info not found, create an empty mgmt vlanconfig: %v", err)
+		if err := h.createMgmtVlanConfig(&networkv1.VlanConfig{ObjectMeta: metav1.ObjectMeta{
+			Name: utils.GetMgmtVlanConfigName(h.nodeName),
+		}}); err != nil {
+			return fmt.Errorf("failed to create empty mgmt vlanconfig: %w", err)
+		}
+		return nil
+	}
+
+	return h.createOrUpdateMgmtVlanConfig(mgmtInfo)
+}
+
+func (h *Handler) createOrUpdateMgmtVlanConfig(mgmtInfo *utils.MgmtBondInterfaceInfo) error {
+	name := utils.GetMgmtVlanConfigName(h.nodeName)
+
+	existing, err := h.vcCache.Get(name)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return err
+	}
+
+	// Resource does not exist -> create using mgmtInfo
+	if apierrors.IsNotFound(err) {
+		vc := &networkv1.VlanConfig{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: name,
+				Labels: map[string]string{
+					utils.KeyClusterNetworkLabel: utils.ManagementClusterNetworkName,
+				},
+			},
+			Spec: networkv1.VlanConfigSpec{
+				ClusterNetwork: utils.ManagementClusterNetworkName,
+				Uplink: networkv1.Uplink{
+					BondOptions: &networkv1.BondOptions{
+						Mode:   networkv1.BondMode(mgmtInfo.BondMode),
+						Miimon: -1,
+					},
+					LinkAttrs: &networkv1.LinkAttrs{
+						MTU:    mgmtInfo.MTU,
+						TxQLen: -1,
+					},
+					NICs: mgmtInfo.SlaveNames,
+				},
+			},
+		}
+
+		return h.createMgmtVlanConfig(vc)
+	}
+
+	// Resource exists -> update if required
+	vcCopy := existing.DeepCopy()
+
+	if !utils.BuildUpdatedVlanConfig(vcCopy, mgmtInfo) {
+		return nil
+	}
+
+	_, err = h.vcClient.Update(vcCopy)
+	if err != nil {
+		return err
+	}
+
+	logrus.Infof("updated VlanConfig %s with MTU=%d BondMode=%s NICs=%v", vcCopy.Name, vcCopy.Spec.Uplink.LinkAttrs.MTU, vcCopy.Spec.Uplink.BondOptions.Mode, vcCopy.Spec.Uplink.NICs)
+
+	return nil
 }
--- a/pkg/controller/manager/node/controller.go
+++ b/pkg/controller/manager/node/controller.go
@@ -9,6 +9,7 @@ import (
 	mapset "github.com/deckarep/golang-set/v2"
 	ctlcorev1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 
@@ -83,6 +84,10 @@ func (h Handler) OnChange(_ string, node *corev1.Node) (*corev1.Node, error) {
 		return nil, err
 	}
 	for _, vc := range vcs {
+		if vc.Spec.ClusterNetwork == utils.ManagementClusterNetworkName {
+			// skip vlan config with management cluster network
+			continue
+		}
 		if err := h.updateMatchedNodeAnnotation(vc, node); err != nil {
 			return nil, fmt.Errorf("failed to update matched node annotation, vc: %s, node: %s, error: %w",
 				vc.Name, node.Name, err)
@@ -242,9 +247,27 @@ func (h Handler) removeNodeFromVlanConfig(nodeName string) error {
 	}
 
 	for _, vc := range vcs {
+		if vc.Spec.ClusterNetwork == utils.ManagementClusterNetworkName {
+			if vc.Name == utils.GetMgmtVlanConfigName(nodeName) {
+				if err := h.removeMgmtVlanConfig(nodeName); err != nil {
+					return err
+				}
+			}
+			continue
+		}
 		if err := h.removeNodeFromOneVlanConfig(vc, nodeName); err != nil {
 			return err
 		}
+	}
+
+	return nil
+}
+
+func (h Handler) removeMgmtVlanConfig(nodeName string) error {
+	vcName := utils.GetMgmtVlanConfigName(nodeName)
+
+	if err := h.vcClient.Delete(vcName, &metav1.DeleteOptions{}); err != nil && !apierrors.IsNotFound(err) {
+		return err
 	}
 
 	return nil
--- a/pkg/controller/manager/vlanconfig/controller.go
+++ b/pkg/controller/manager/vlanconfig/controller.go
@@ -24,6 +24,7 @@ type Handler struct {
 	cnCache  ctlnetworkv1.ClusterNetworkCache
 	vsCache  ctlnetworkv1.VlanStatusCache
 	vcCache  ctlnetworkv1.VlanConfigCache
+	vcClient ctlnetworkv1.VlanConfigClient
 }
 
 func Register(ctx context.Context, management *config.Management) error {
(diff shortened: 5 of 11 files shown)


pkg/controller/agent/linkmonitor/controller.go
```diff
@@ -207,3 +217,72 @@ func (h Handler) syncLinkStatus(lm *networkv1.LinkMonitor) error {
...
+	bondModeChanged := vc.Spec.Uplink.BondOptions == nil || string(vc.Spec.Uplink.BondOptions.Mode) != mgmtIntfInfo.BondMode
+
+	if mtuChanged {
+		vcCopy.Spec.Uplink.LinkAttrs.MTU = mgmtIntfInfo.MTU
+	}
+	if bondModeChanged {
+		vcCopy.Spec.Uplink.BondOptions.Mode = networkv1.BondMode(mgmtIntfInfo.BondMode)
+	}
+
+	slavesChanged := !compareStringSlices(vc.Spec.Uplink.NICs, mgmtIntfInfo.SlaveNames)
```
Yu-Jack: We can use `slices.Equal`.

pkg/controller/manager/vlanconfig/controller.go
```diff
@@ -46,6 +51,57 @@ func Register(ctx context.Context, management *config.Management) error {
...
+				LinkAttrs: &networkv1.LinkAttrs{
+					MTU:    mgmtIntfInfo.MTU,
+					TxQLen: -1,
+				},
+
+				NICs: func() []string {
+					nics := make([]string, 0, len(mgmtIntfInfo.SlaveNames))
+					nics = append(nics, mgmtIntfInfo.SlaveNames...)
+					return nics
+				}(),
```
Yu-Jack: I don't see any places where we manipulate `mgmtIntfInfo.SlaveNames`, so it should be able to just use `NICs: mgmtIntfInfo.SlaveNames`. Or if you do this for slice isolation, it's also okay.

pkg/controller/manager/vlanconfig/controller.go
```diff
@@ -46,6 +51,57 @@ func Register(ctx context.Context, management *config.Management) error {
 	return nil
 }
 
+// CreateMgmtVlanConfig checks mgmt-bo, gathers info, and creates a VlanConfig resource.
+func (h *Handler) CreateMgmtVlanConfig() error {
```
w13915984028: it looks to be the responsibility of `agent` controllers to do this work:

on a cluster, each node could have it's own NICs to be used as mgmt uplink

vlanconfig:node 1:1 mapping should be better

pkg/controller/agent/linkmonitor/controller.go
```diff
@@ -207,3 +217,72 @@ func (h Handler) syncLinkStatus(lm *networkv1.LinkMonitor) error {
 
 	return h.updateStatus(lm, linkStatusList)
 }
+
+// update mgmt vlanconfig when mgmt cluster link params change
+func (h Handler) updateMgmtVlanConfig() error {
```
w13915984028, line 223: on a non-unified network topologies cluster, agent should only update it's vlanconfig, not others

w13915984028 (commented):
for mgmt network, the vlanconfig:node is better to have 1:1 mapping, as different node could have different nics for mgmt usage, though other params are mostly same

it becomes the agent controller's responsibility to sync information from `ip link`, which is set on the bootstrap stage via `nm`, to `vlanconfig`

the `manager` controller removes those stale mgmt vlanconfig after a node has been removed from the cluster

user can only view the mgmt vlanconfig, but can't modify them, to avoid race with `nm`

pkg/controller/manager/vlanconfig/controller.go
rrajendran17: Yes right,  I was under the assumption that interfaces on all nodes are same. But creating vlan config per node is the right way to do it.Thanks for the insight. I have updated the changes.

pkg/controller/agent/linkmonitor/controller.go
rrajendran17: slices.Equal returns false even for same set in different order, and it  will require sorting of old and new sets before comparison.This utility returns true for same unordered sets and faster. Better to keep this.

pkg/controller/agent/vlanconfig/controller.go
```diff
@@ -462,3 +468,99 @@ func (h Handler) removeNodeLabel(vs *networkv1.VlanStatus) error {
...
+
+// CreateOrUpdateMgmtVlanConfig checks mgmt-bo, gathers info, and creates or updates a VlanConfig resource.
+func (h *Handler) CreateOrUpdateMgmtVlanConfig() error {
+	mgmtInfo, err := utils.GetMgmtBondInfo()
+	if err != nil {
+		logrus.Warnf("mgmt interface info not found, skipping mgmt-vlanconfig creation/update: %v", err)
+		return nil
+	}
+
+	mgmtVlanConfigName := utils.MgmtVlanConfig + "-" + h.nodeName
```
w13915984028: might be necessary to use below function for all related names in  this PR:
"github.com/rancher/wrangler/v3/pkg/name"
`name.SafeConcatName(utils.MgmtVlanConfig, h.nodeName)`

pkg/utils/bondInterface.go
```diff
@@ -0,0 +1,100 @@
...
+	changed := false
+
+	if vc.Spec.Uplink.LinkAttrs == nil {
+		vc.Spec.Uplink.LinkAttrs = &networkv1.LinkAttrs{TxQLen: -1}
+	}
+	if vc.Spec.Uplink.BondOptions == nil {
+		vc.Spec.Uplink.BondOptions = &networkv1.BondOptions{Miimon: -1}
+	}
+
+	if vc.Spec.Uplink.LinkAttrs.MTU != mgmtIntfInfo.MTU {
```
w13915984028, line 67: MTU is a bit special, 0, 1500 are equal

pkg/utils/bondInterface.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/vishvananda/netlink"
+)
+
+type MgmtBondInterfaceInfo struct {
+	MTU        int
+	BondMode   string
+	SlaveNames []string
+}
+
+func GetMgmtBondInfo() (MgmtBondInterfaceInfo, error) {
```
w13915984028: Is below a bit better ? error case just return `nil, err`

`
func GetMgmtBondInfo() (*MgmtBondInterfaceInfo, error) 
`

pkg/controller/manager/node/controller.go
```diff
@@ -250,11 +254,34 @@ func (h Handler) removeNodeFromVlanConfig(nodeName string) error {
 	return nil
 }
 
+func (h Handler) removeMgmtVlanConfig(nodeName string) error {
+	vcName := utils.MgmtVlanConfig + "-" + nodeName
+	vc, err := h.vcCache.Get(vcName)
+	if err != nil {
+		return err
+	}
+	if vc == nil {
```
w13915984028: normally `vc == nil` comes with `IsNotFound` error; 

Put L263 inside L261?

w13915984028 (commented):
Please check those points, thanks.

pkg/controller/agent/vlanconfig/controller.go
```diff
@@ -462,3 +468,99 @@ func (h Handler) removeNodeLabel(vs *networkv1.VlanStatus) error {
...
+	vc := &networkv1.VlanConfig{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: mgmtVlanConfigName,
+			Labels: map[string]string{
+				utils.KeyClusterNetworkLabel: utils.ManagementClusterNetworkName,
+			},
+		},
+		Spec: networkv1.VlanConfigSpec{
+			ClusterNetwork: utils.ManagementClusterNetworkName,
+			Uplink: networkv1.Uplink{
```
w13915984028: where is
```
	NodeSelector   map[string]string `json:"nodeSelector,omitempty"`
```

when vlanconfig:node are 1:1 ?

pkg/controller/agent/vlanconfig/controller.go
rrajendran17: Since its important to retain the node names in the mgmt-vlanconfig and it cannot be guaranteed to lesser than 64, above SafeConcatName can truncate the name in worst cases(long names). So will retain the existing way.

pkg/utils/bondInterface.go
rrajendran17: Since MTU is fetched from ip command on interface and not manipulated inside controller, it cannot be 0.

pkg/utils/bondInterface.go
rrajendran17: Since the caller just checks for the error, the existing one should ok imo.

pkg/controller/agent/vlanconfig/controller.go
rrajendran17: We will not need an explicit node selector, agent on each node creates a vlanconfig (mgmt-vlanconfig-nodename) and number of vlanconfigs will be equal to number of nodes in the cluster for mgmt cluster network.Link monitor per agent handles changes to its respective mgmt-vlanconfig-nodename and node removal removes only the vlanconfig in that nodename.

pkg/controller/manager/node/controller.go
```diff
@@ -250,11 +255,34 @@ func (h Handler) removeNodeFromVlanConfig(nodeName string) error {
...
+
+	if err := h.vcClient.Delete(vcName, &metav1.DeleteOptions{}); err != nil {
+		return err
+	}
+
+	return nil
+}
+
 func (h Handler) removeNodeFromOneVlanConfig(vc *networkv1.VlanConfig, nodeName string) error {
 	if vc.Annotations == nil || vc.Annotations[utils.KeyMatchedNodes] == "" {
```
w13915984028, line 277: when the created `mgmt` vc has no `nodeSelector`, the annotation `vc.Annotations[utils.KeyMatchedNodes]` is empty, L281 might have no chance to be called

Or set this annotation directly from

https://github.com/harvester/network-controller-harvester/pull/276/changes#diff-863291daa63e82665b9a39001e756b427f4cc8203fbc97b4f2750173310e50e5R495

pkg/utils/bondInterface.go
w13915984028: the slight difference between
```
return MgmtBondInterfaceInfo{}, err
```
and
```
return nil, err
```

with the first one, on any of it's errors, an empty `MgmtBondInterfaceInfo{}` need to be constructed and returned, but caller does not use it at all

pkg/controller/agent/vlanconfig/controller.go
w13915984028: I guess this goes to a wrong direction

The `vlanconfig`'s name is constructed via two elements, it by nature can exceed limits

when use a helper function to generate the safe name, this helper function should be used on all the names processing of this PR; and, if node name is still required, it could be added to `label`, and later use the `label` to select it

w13915984028 (commented):
Please check those open topics, thanks.

pkg/controller/manager/node/controller.go
w13915984028, line 258: the `remove` logic could be added on this existing for loop or `removeNodeFromOneVlanConfig`, where both `vc` and `nodeName` are already on hand

https://github.com/harvester/network-controller-harvester/blob/116ebb6ce048bd2c587c99be556789baa8873447/pkg/controller/manager/node/controller.go#L238-L248

if this is a mgmt vc, then check the `vc.Name` or `vc.Label[node-name]` with node, and delete it when matches;  it does not require to query the vc again on current PR

pkg/controller/agent/vlanconfig/controller.go
rrajendran17: I do not see any explicit k8s doc mentioning about 64 chars on custom resource names.But I see we use safeConcatName in some of the places in harvester.Can it be because labels are restricted to 64 chars and we wanted some of the resource names to be within limit to be used in labels for future references ? 
Also using the node name in vlanconfig label will also be not an option as node name does not have limits but label value has 64 chars upper limit.
I still feel using a vlanconfig name without any upper limit should be ok. Please correct me if there are some reasons which I am missing.

pkg/controller/manager/node/controller.go
rrajendran17: yes, that's better

pkg/controller/agent/vlanconfig/controller.go
w13915984028: AI generated.

image

From any time, when call `name.SafeConcatName(utils.MgmtVlanConfig, h.nodeName)`, the result is stable, predictable. It helps to avoid triggering any potential traps.

```
// SafeConcatName concatenates the given strings and ensures the returned name is under 64 characters
// by cutting the string off at 57 characters and setting the last 6 with an encoded version of the concatenated string.
func SafeConcatName(name ...string) string {
	fullPath := strings.Join(name, "-")
	if len(fullPath) < 64 {
		return fullPath
	}
	digest := sha256.Sum256([]byte(fullPath))
	// since we cut the string in the middle, the last char may not be compatible with what is expected in k8s
	// we are checking and if necessary removing the last char
	c := fullPath[56]
	if 'a' <= c && c <= 'z' || '0' <= c && c <= '9' {
		return fullPath[0:57] + "-" + hex.EncodeToString(digest[0:])[0:5]
	}
}
```

rrajendran17: Yeah I understand the differences and any CR name could be upto 253 chars and restricting it to 64 will not be right and also this is an internally managed object and not used by any other resources, so its safe to append it with the node name. To be even safer, may be I will change the mgmt vlanconfig name to be name of the node itself instead of prefix mgmt-vlanconfig.

pkg/controller/agent/vlanconfig/controller.go
w13915984028: Now we add many `mgmt` related checks to shortcut the normal vlanconfig processing.

It will be even robust to add the `NodeSelector` to align with the original design, VlanConfig configs `NodeSelector` and mutator/controller updates the matched node name to label/annotation, controller reads the label/annotation to do further processing.

This avoids the potential breaking on some unnoticed corner cases. thanks.

Also refer doc https://docs.harvesterhci.io/v1.8/networking/index#create-a-new-cluster-network step 6. User can view all vlanconfig objects. Keep same logic also avoids to explain to user that `this is mgmt network vlanconfig, it is special ...`;  the mgmt vlanconfig can even be a sample when user creates customized cluster networks.

pkg/controller/agent/vlanconfig/controller.go
w13915984028: Use node name is straightforward and dismiss those concerns.

pkg/controller/agent/vlanconfig/controller.go
rrajendran17: How node selectors work with vlanconfig is, when user selects all nodes from UI, vlanconfig mutator adds all node names in the cluster to the key matched nodes annotation. But with node selectors (key:value), mutator adds only the nodes matching the key:value to the key matched nodes annotation. So it will not be required here for mgmt vlanconfig and I do not see a purpose.

other vlanconfigs needs node selectors(keyMatchedNode annotation) because they have to match on a particular node to create uplink etc on that node. In case of mgmt vlanconfig, that is not necessary and not all checks are based on KeyMatchedAnnotation. 

Even if we add matched nodes per mgmt vlanconfig with node selectors we will still need those checks to skip processing for mgmt cluster network as vlanconfig changes are handled based on matched nodes annotation. https://github.com/harvester/network-controller-harvester/blob/116ebb6ce048bd2c587c99be556789baa8873447/pkg/controller/agent/vlanconfig/controller.go#L106.

I have the UI changes also here https://github.com/harvester/harvester-ui-extension/pull/927, the UI shows this vlanconfig from mgmt cluster network as state Managed different from other vlanconfigs and users are allowed only to view.

pkg/utils/bondInterface.go
```diff
@@ -0,0 +1,100 @@
...
+
+type MgmtBondInterfaceInfo struct {
+	MTU        int
+	BondMode   string
+	SlaveNames []string
+}
+
+func GetMgmtBondInfo() (*MgmtBondInterfaceInfo, error) {
+	// Find the mgmt-bo bond interface
+	mgmtBo, err := netlink.LinkByName("mgmt-bo")
```
w13915984028: add a const like https://github.com/harvester/network-controller-harvester/blob/116ebb6ce048bd2c587c99be556789baa8873447/pkg/utils/bridge.go#L24 for common usage

pkg/utils/bondInterface.go
```diff
@@ -0,0 +1,100 @@
...
+		return nil, err
+	}
+	var slaveNames []string
+	for _, l := range links {
+		if l.Attrs().MasterIndex == mgmtBo.Attrs().Index {
+			slaveNames = append(slaveNames, l.Attrs().Name)
+		}
+	}
+
+	bond, ok := mgmtBo.(*netlink.Bond)
```
w13915984028: should this check be before L24?

pkg/controller/agent/linkmonitor/controller.go
```diff
@@ -207,3 +218,35 @@ func (h Handler) syncLinkStatus(lm *networkv1.LinkMonitor) error {
...
+	mgmtVlanConfigName := utils.MgmtVlanConfig + "-" + h.nodeName
```
w13915984028: no matter using safename or not, it deserves to have a helper function for this name, there are 4+ times of similar code

pkg/controller/agent/vlanconfig/controller.go
```diff
@@ -462,3 +468,99 @@ func (h Handler) removeNodeLabel(vs *networkv1.VlanStatus) error {
 func (h Handler) statusName(clusterNetwork string) string {
 	return utils.Name("", clusterNetwork, h.nodeName)
 }
+
+// CreateOrUpdateMgmtVlanConfig checks mgmt-bo, gathers info, and creates or updates a VlanConfig resource.
+func (h *Handler) CreateOrUpdateMgmtVlanConfig() error {
+	mgmtInfo, err := utils.GetMgmtBondInfo()
+	if err != nil {
+		logrus.Warnf("mgmt interface info not found, skipping mgmt-vlanconfig creation/update: %v", err)
```
w13915984028: it is a bit risky to skip this error;  maybe we should create a dummy object (with no bond-slave) in this case (if it is not existing)

w13915984028 (commented):
A few open points, thanks.

pkg/utils/mgmt.go
```diff
@@ -1,7 +1,30 @@
 package utils
 
-const ManagementClusterNetworkName = "mgmt"
+import (
+	"github.com/harvester/webhook/pkg/server/admission"
+)
+
+const (
+	ManagementClusterNetworkName = "mgmt"
+	MgmtVlanConfigSuffix         = "mgmt-vlanconfig-"
```
w13915984028: this looks to be a prefix

pkg/controller/agent/vlanconfig/controller.go
```diff
@@ -462,3 +468,112 @@ func (h Handler) removeNodeLabel(vs *networkv1.VlanStatus) error {
...
+}
+
+// CreateOrUpdateMgmtVlanConfig checks mgmt-bo, gathers info, and creates or updates a VlanConfig resource.
+func (h *Handler) CreateOrUpdateMgmtVlanConfig() error {
+	mgmtInfo, err := utils.GetMgmtBondInfo()
+	if err != nil {
+		logrus.Warnf("mgmt interface info not found, create an empty mgmt vlanconfig: %v", err)
+		if err := h.createMgmtVlanConfig(&networkv1.VlanConfig{ObjectMeta: metav1.ObjectMeta{
+			Name: utils.GetMgmtVlanConfigName(h.nodeName),
+		}}); err != nil {
```
w13915984028, line 495: skip isAlreadyExisting error

w13915984028 (dismissed):
LGTM, thanks.

pkg/utils/bondInterface.go
```diff
@@ -0,0 +1,100 @@
...
+		changed = true
+	}
+
+	if vc.Spec.Uplink.BondOptions.Mode != networkv1.BondMode(mgmtIntfInfo.BondMode) {
+		vc.Spec.Uplink.BondOptions.Mode = networkv1.BondMode(mgmtIntfInfo.BondMode)
+		changed = true
+	}
+
+	if !compareStringSlices(vc.Spec.Uplink.NICs, mgmtIntfInfo.SlaveNames) {
+		vc.Spec.Uplink.NICs = append([]string{}, mgmtIntfInfo.SlaveNames...)
```
ibrokethecloud: why not just assign mgmtInfo.SlaveNames directly?

pkg/controller/agent/vlanconfig/controller.go
```diff
@@ -462,3 +468,112 @@ func (h Handler) removeNodeLabel(vs *networkv1.VlanStatus) error {
...
+			return fmt.Errorf("failed to create empty mgmt vlanconfig: %w", err)
+		}
+		return nil
+	}
+
+	mgmtVlanConfigName := utils.GetMgmtVlanConfigName(h.nodeName)
+
+	// Try update first.
+	found, err := h.updateMgmtVlanConfigIfExists(mgmtInfo)
```
ibrokethecloud: this could be simplified if the method was say changed to `createOrUpdateMgmtVlanConfigIfExists`
The method could create the vlanconfig object if apierrors.IsNotFound(err) is true, ensureing all the logic for creation / updation is one method which just returns `err` and the overall logic in `CreateOrUpdateMgmtVlanConfig` would get simplified

pkg/controller/manager/node/controller.go
```diff
@@ -250,6 +261,16 @@ func (h Handler) removeNodeFromVlanConfig(nodeName string) error {
 	return nil
 }
 
+func (h Handler) removeMgmtVlanConfig(nodeName string) error {
+	vcName := utils.GetMgmtVlanConfigName(nodeName)
+
+	if err := h.vcClient.Delete(vcName, &metav1.DeleteOptions{}); err != nil {
```
ibrokethecloud: we could add check to ignore apierrors.IsNotFound(err) to avoid additional reconciles or conflicts if for some reason the vlanconfig is not found

pkg/controller/agent/vlanconfig/controller.go
rrajendran17: createMgmtVlanConfig checks for alreadyexists

rrajendran17:
@w13915984028 @ibrokethecloud The PR is ready after addressing the comments, Thanks

cmd/webhook/main.go
```diff
@@ -87,7 +87,7 @@ func main() {
 			Name:        "controller-user",
 			EnvVar:      "CONTROLLER_USER_NAME",
 			Destination: &options.ControllerUsername,
-			Value:       "harvester-load-balancer-webhook",
+			Value:       "system:serviceaccount:harvester-system:harvester-network-controller",
```
w13915984028, line 90: there is `harvester-network-webhook` service account, should we use it?

```
apiVersion: v1
kind: ServiceAccount
metadata:
  name: harvester-network-webhook
```

pkg/controller/manager/node/controller.go
```diff
@@ -242,6 +247,13 @@ func (h Handler) removeNodeFromVlanConfig(nodeName string) error {
 	}
 
 	for _, vc := range vcs {
+		if vc.Spec.ClusterNetwork == utils.ManagementClusterNetworkName {
```
w13915984028, line 250: as the vcs is an everything list, the for loop needs to check two conditions
`if vc.Spec.ClusterNetwork == utils.ManagementClusterNetworkName  && vc.Name == utils.GetMgmtVlanConfigName(nodeName) {
}`

before call delete;  currently it will try to delete a same node multi times

pkg/utils/bondInterface.go
```diff
@@ -0,0 +1,100 @@
...
+	if !compareStringSlices(vc.Spec.Uplink.NICs, mgmtIntfInfo.SlaveNames) {
+		vc.Spec.Uplink.NICs = mgmtIntfInfo.SlaveNames
+		changed = true
+	}
+
+	return changed
+}
+
+// Helper to compare two string slices
+func compareStringSlices(a, b []string) bool {
```
w13915984028, line 86: there are two public functions to help:

(1) on the generate part, use `slices.Sort(slaveNames)` https://github.com/harvester/network-controller-harvester/pull/276/changes#diff-63b602ece5897eb4272db4e551ce5ba857eba5040ba3882f2b6cd963acdc05adR47

(2) on compare, use 

// Returns true if lengths match and elements match index-for-index
slices.Equal(sliceA, sliceB)

or replace it on https://github.com/harvester/network-controller-harvester/pull/276/changes#diff-63b602ece5897eb4272db4e551ce5ba857eba5040ba3882f2b6cd963acdc05adR77 directly

w13915984028 (commented):
Please check, thanks.

pkg/utils/bondInterface.go
rrajendran17: Yeah I thought since the slice is going to be minimal, we could avoid using sort function call and use a own network utility.

cmd/webhook/main.go
rrajendran17: ControllerUsername is based on what is received from the req.UserInfo.Username and we receive "system:serviceaccount:harvester-system:harvester-network-controller"

w13915984028: Got it. The field looks not used before, but this PR check the req.

pkg/controller/manager/node/controller.go
```diff
@@ -242,6 +247,13 @@ func (h Handler) removeNodeFromVlanConfig(nodeName string) error {
 	}
 
 	for _, vc := range vcs {
+		if vc.Spec.ClusterNetwork == utils.ManagementClusterNetworkName && vc.Name == utils.GetMgmtVlanConfigName(nodeName) {
```
w13915984028: it seems mgmt vcs need to be processed as following:

```
		if vc.Spec.ClusterNetwork == utils.ManagementClusterNetworkName {
			if vc.Name == utils.GetMgmtVlanConfigName(nodeName) {
				// if the vlan config is for management cluster network, delete the vlan config directly because management cluster network only matches one node
				if err := h.removeMgmtVlanConfig(nodeName); err != nil {
					return err
				}
			}
			continue
		}
```

w13915984028 (dismissed):
LGTM with a last open point, thanks.

pkg/controller/manager/node/controller.go
rrajendran17: Right, my bad missed it, Thanks

w13915984028 (dismissed):
LGTM, thanks.

w13915984028 (approved):
LGTM, thanks.

rrajendran17:
@mergify backport v1.9

PR #376, opened 2026-08-16
deps: update insomniacslk/dhcp to reject malformed short IPv4 frames

thc1006:
This repo pins insomniacslk/dhcp at `v0.0.0-20260603135910-a415979eb11e`, before insomniacslk/dhcp#583, and this moves it to the merge commit.

Before #583, `BroadcastRawUDPConn.ReadFrom` (`dhcpv4/nclient4/conn_unix.go`) subtracted the 8-byte UDP header from the IPv4 payload length without checking the payload was at least that long, so a reply with a smaller declared payload gave a negative DHCP length and panicked on a bad slice bound.

Only one of the two DHCP paths here reaches that code. `sendDiscoverMessage` in `pkg/helper/dhcp.go` builds its client with `nclient4.New(iface)` (default raw broadcast) and `DiscoverOffer` reads through the affected `ReadFrom`; `NewLeaseManager` in `pkg/controller/agent/hostnetworkconfig/leasemanager.go` uses the same default client via `client.Request`. `sendInformMessage` passes `nclient4.WithUnicast`, so its replies come back on a normal UDP socket and it isn't affected.

I wrote the upstream fix. I'm not claiming a specific remote exploit against Harvester; a crafted reply would have to come from whatever answers DHCP on the VLAN. `go build ./cmd/... ./pkg/...`, `go vet`, and `go test ./pkg/helper/... ./pkg/controller/agent/hostnetworkconfig/...` pass, and the vendor tree was refreshed with `go mod tidy` and `go mod vendor`. One caveat so reviewers aren't surprised: `go build ./...` fails on the repo-root `main.go` both before and after this change, since that file only carries `//go:generate` directives with no `func main`. `scripts/build` compiles the three cmd binaries directly, so nothing new breaks here.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -75,7 +75,7 @@ require (
 	github.com/deckarep/golang-set/v2 v2.9.0
 	github.com/go-ping/ping v0.0.0-20211014180314-6e2b003bffdd
 	github.com/harvester/webhook v0.1.5
-	github.com/insomniacslk/dhcp v0.0.0-20260603135910-a415979eb11e
+	github.com/insomniacslk/dhcp v0.0.0-20260719225207-c76316d4aa82
 	github.com/k8snetworkplumbingwg/network-attachment-definition-client v1.7.7
 	github.com/kubeovn/kube-ovn v1.13.13
 	github.com/rancher/lasso v0.2.2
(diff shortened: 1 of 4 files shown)


w13915984028 (approved):
LGTM, thanks.

PR #382, opened 2026-08-27
[WIP]Optimize HostNetworkConfig OnChange

w13915984028:
**Problem:**

1. HostNetworkConfig OnChange is shortcutted randomly.

```
if intfExists {
    // BUG: Shortcuts immediately without checking if:
    // 1. hnc.Spec.HostIPs changed to a new IP for this node
    // 2. hnc.Spec.Mode changed from Static to DHCP (or vice versa)
    // 3. the nic was not up on last try
    // 4. ip was not set
    // 5. dhcp failed in the middle

    return hnc, nil
}
```

2. Additional observation: due to the `single object with multi node status` model, it could be affected by any flaky node.

```
[Flaky Node-A] (Interface Flaps / Errors out)
      │
      ├──> Patches status.nodeStatus["Node-A"] (Error state)
      │
      ▼
[K8s API Server] (Emits WATCH event for HostNetworkConfig v123)
      │
      ├───► Broadcasts WATCH event to Node-A
      ├───► Broadcasts WATCH event to Node-B
      ├───► Broadcasts WATCH event to Node-C
      └───► Broadcasts WATCH event to Node-N
      │
      ▼
[All Healthy Nodes (B...N)]
      │
      ├──> Receives WATCH event for Node-A's status change
      ├──> Enters OnChange() blindly (No DeepEqual Guard)
      ├──> Re-executes Netlink / DHCP / IP commands
      ├──> Re-patches status.nodeStatus["Node-X"] to API Server
      │
      ▼
[API Server Storm] (N new WATCH events generated!)
      │
      └───► Loop amplifies exponentially until backoff limit is hit
```

**Solution:**

1. Ensure the OnChange is idempotent
2. Add a check between local status and CRD status

**Related Issue:**

https://github.com/harvester/harvester/issues/11519

**Test plan:**

PR #383, opened 2026-08-28
fix(webhook): skip self-comparison in validateHostNetworkConfig during updates

w13915984028:
**Problem:**

During an Update operation on an existing HostNetworkConfig (HNC), the admission webhook's validation logic (validateHostNetworkConfig) iterates over all cached HNC objects to check for duplicate underlay configurations and (ClusterNetwork, VlanID) collisions.

Because the cache contains the pre-update version of the CR being modified, the loop compares the updated CR against itself (hostnetworkconfig.Name == newhnc.Name). This results in false-positive validation errors:

Updating non-immutable fields on an existing Underlay: true HNC is erroneously rejected because hostnetworkconfig.Spec.Underlay && newhnc.Spec.Underlay evaluates to true.

Any general update validation encounters self-collision on (ClusterNetwork, VlanID).

**Solution:**

Add an explicit name check if hostnetworkconfig.Name == newhnc.Name { continue } inside validateHostNetworkConfig to skip self-comparison during update operations.

**Related Issue:**

https://github.com/harvester/harvester/issues/11603

observed while working on the main PR https://github.com/harvester/network-controller-harvester/pull/382

**Test plan:**

Original issue:
image

(1) Create an underlay mode hostnetworkconfig, then change the host ip, it is denied on old webhook, but allowed after the fix.
image

(2) Change the IP.
```
apiVersion: v1
items:
- apiVersion: network.harvesterhci.io/v1beta1
  kind: HostNetworkConfig
  metadata:
    creationTimestamp: "2026-09-08T10:09:51Z"
    finalizers:
    - wrangler.cattle.io/harvester-network-hostnetworkconfig-controller
    generation: 4
    name: hnc1
    resourceVersion: "718905"
    uid: ffa45d51-624f-43df-80bd-0da6a2ce05ef
  spec:
    clusterNetwork: mgmt
    ips:
      harv31: 192.168.3.11/24  // webhook does not block this change
    mode: static
    underlay: true
    vlanID: 3
  status:
    nodeStatus:
      harv31:
        clusterNetwork: mgmt
        conditions:
        - message: ""
          status: "True"
          type: ready
        mode: static
        vlanID: 3
kind: List
metadata:
  resourceVersion: ""
```

Note: The IP out-of-sync is booked on https://github.com/harvester/harvester/issues/11618. It will be sovled via PR https://github.com/harvester/network-controller-harvester/pull/395 and validated per issue 11618.

IP was not updated on node yet.
```
79: mgmt-br.3@mgmt-br: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default 
    link/ether 52:54:00:c1:9d:05 brd ff:ff:ff:ff:ff:ff
    inet 192.168.3.10/24 brd 192.168.3.255 scope global mgmt-br.3
       valid_lft forever preferred_lft forever
```

### DIFF
--- a/pkg/webhook/hostnetworkconfig/validator.go
+++ b/pkg/webhook/hostnetworkconfig/validator.go
@@ -302,6 +302,11 @@ func (v *Validator) validateHostNetworkConfig(newhnc *networkv1.HostNetworkConfi
 			continue
 		}
 
+		// Skip comparing the CR against itself during updates
+		if hostnetworkconfig.Name == newhnc.Name {
+			continue
+		}
+
 		//no more than one underlay exists for overlay networks. If user has to change underlay setting, old underlay has to be disabled first.
 		if hostnetworkconfig.Spec.Underlay && newhnc.Spec.Underlay {
 			return fmt.Errorf("hostnetworkconfig %s already exists for cluster network %s vlanid %d with underlay enabled", hostnetworkconfig.Name, newhnc.Spec.ClusterNetwork, hostnetworkconfig.Spec.VlanID)
@@ -311,8 +316,10 @@ func (v *Validator) validateHostNetworkConfig(newhnc *networkv1.HostNetworkConfi
 			continue
 		}
 
+		// Enforce VLAN ID uniqueness on creation only to avoid blocking updates on existing duplicate states
 		if op == ValidateCreate && hostnetworkconfig.Spec.VlanID == newhnc.Spec.VlanID {
-			return fmt.Errorf("hostnetworkconfig %s already exists for cluster network %s with vlanID %d", hostnetworkconfig.Name, hostnetworkconfig.Spec.ClusterNetwork, hostnetworkconfig.Spec.VlanID)
+			return fmt.Errorf("hostnetworkconfig %s already exists for cluster network %s with vlanID %d",
+				hostnetworkconfig.Name, hostnetworkconfig.Spec.ClusterNetwork, hostnetworkconfig.Spec.VlanID)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


rrajendran17 (approved):
LGTM,Thanks

martindekov (commented):
Just a general question, this would mean that if we update we allow all changes in the new cr without validation if names are the same (same cr), is that right? 

Don't we also allow breaking changes to the cr that way as a side effect to skipping validation in that scenario?

w13915984028:
> Just a general question, this would mean that if we update we allow all changes in the new cr without validation if names are the same (same cr), is that right?
> 
> Don't we also allow breaking changes to the cr that way as a side effect to skipping validation in that scenario?

A good question.

From the webhook, L130, L135 checks those immutable fields of current object;
https://github.com/harvester/network-controller-harvester/blob/f383a15a2c09373edbe9f48e0027b87b224abf60/pkg/webhook/hostnetworkconfig/validator.go#L125-L141

L141 loop checks the current object upon all other objects to avoid some global limitations be violated. By natue, it should avoid to compare self, e.g., below code will deny change on self (any allow to be changed spec fields) when `Spec.Underlay` is true.


```
		//no more than one underlay exists for overlay networks. If user has to change underlay setting, old underlay has to be disabled first.
		if hostnetworkconfig.Spec.Underlay && newhnc.Spec.Underlay {
			return fmt.Errorf("hostnetworkconfig %s already exists for cluster network %s vlanid %d with underlay enabled", hostnetworkconfig.Name, newhnc.Spec.ClusterNetwork, hostnetworkconfig.Spec.VlanID)
		}
```

martindekov (approved):
Thanks for the context Jian, from change itself I couldn't infer that we already validate the `newhnc` before the continue call in this PR. Allowing potential breaking changes to the CR so slip was my only concern and the context resolved this. LGTM!

PR #404, opened 2026-09-11
[CI] Add validate check to avoid import harvester accidentally

w13915984028:
**Problem:**

Harvester-network-controller ever looped importing harvester, and the loop dependency was decoupled before via PR https://github.com/harvester/network-controller-harvester/pull/217.

Now, there is a single direction dependency: harvester imports network-controller.

https://github.com/harvester/harvester/blob/8012209e56a610d73bfe453bf59283d496b5dcee/go.mod#L13

**Solution:**

This PR adds a check on `make validate` to ensure the depency is not added again accidentally.

**Related Issue:**

**Test plan:**

Not required.

`make validate` runs smoothly

### DIFF
--- a/scripts/validate
+++ b/scripts/validate
@@ -3,6 +3,17 @@ set -e
 
 cd $(dirname $0)/..
 
+# WARNING: Do not import harvester/harvester to prevent circular dependencies
+# (harvester/harvester already depends on this repository).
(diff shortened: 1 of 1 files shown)


martindekov (approved):
LGTM!

w13915984028:
As this is a passive CI check script, merge it now.

PR #416, opened 2026-09-22
Deny HNC mode change

w13915984028:
Controller side cannot clean resources clearly if mode is changed
  Update test code

**Problem:**

1. If HNC mode is changed, it does not clean the previous IP&routes, instead it applies new settings directly.
2. It also has no full context of previous config.

**Solution:**

Deny the change of mode, unless controller side can process this fully.

**Related Issue:**

https://github.com/harvester/harvester/issues/11617

**Test plan:**

1. Create any type of HNC
2. Change the type, it gets denied.

image

image


w13915984028:
@mergify backport v1.9

w13915984028:
@rrajendran17 @Vicente-Cheng Would suggest to backport this PR to v1.9.1, please take a look, thanks.

PR #418, opened 2026-09-22
Check nil object on delete case

w13915984028:
**Problem:**

For webhook `Delete`, it does not check input param, and use the pointer directly.

**Solution:**

Add a check to skip processing the nil pointer in `Delete` case.

**Related Issue:**

https://github.com/harvester/harvester/issues/10759

**Test plan:**


martindekov (approved):
LGTM

